Issue #18, opened 2016-01-15
Add support for 'fix' from eslint rules to automatically fix eslint errors that are reported.


jeromu:
When writing eslint rules, you can have an optional 'fix' that will automatically patch your code.  It would be great for vscode-eslint to support that.

dbaeumer:
Cool idea.

dbaeumer:
This is available in the latest eslint although with only one pass of fixes. For the n pass case a issue exists.

Issue #10, opened 2015-12-04
Display "ESLint" and the triggered rule within the tooltip


LittleBrainz:
Overall, I am very pleased with vscode-eslint. However, when ESLint inappropriately (from my point of view) flags some offending syntax, it is difficult to determine which rule caused the error.

Also, when several linters are active (including the built VS Code IntelliSense), I am not certain which is causing the text to be highlighted.

Would it be possible to format the tooltip along the lines of:

```
ESLint: you have done something stupid (stupid-syntax-rule)
```

The same format would be ideal if included in the VS Code error summary (ctrl-p + !).

Cheers!

egamma:
Now that I've started to work with TSLint I'm starting to feel the same, that is I'd like to see whether are warning comes from TypeScript or TSLint. 

@dbaeumer I suggest we come up with a guideline and then apply it across all the linters.

My preference would be to show the linter/source at the end of the message, i.e.,

`You have done something stupid (stupid-syntax-rule) [ESLint]`

The message is the 'main' thing and it should come first before the linter which is secondary information.

LittleBrainz:
@egamma - I agree.

dbaeumer:
I think we should not mangle this (from an API perspective) into the message. Instead we should enhance the Diagnostic API with a source field and allow linters to fill in that field. This would allow for proper filtering and grouping in the UI side. See https://github.com/Microsoft/vscode/issues/1010 for the discussion on the VS Code side.

egamma:
I was only taking about the rendering.. Fully agree, that this should not be done with string concatenation, but an API.

dbaeumer:
This got addressed a long time ago.

dbaeumer:
Got fixed a long time ago.

PR #19, opened 2016-01-18
Add autofix supports.

mysticatea:
Refs #10 and #18.

Hello, thank you for the great extension!

The purpose of this PR is that adding the feature of ESLint's Autofix.
This PR is adding one command and one option.

- `eslint.executeAutofix` command executes Autofix on the current document.
  autofix-command
- If `eslint.enableAutofixOnSave` option was enabled then it executes Autofix when you saved a document.
  autofix-on-save

I'm happy if this PR is accepted.
And I'm happy to follow reviews and directions.

(I'm a beginner on TypeScript and VSCode extensions.)

Regards.

### DIFF
--- a/eslint/package.json
+++ b/eslint/package.json
@@ -21,7 +21,9 @@
 		"vscode": "^0.10.8"
 	},
 	"activationEvents": [
-		"onLanguage:javascript", "onLanguage:javascriptreact"
+		"onLanguage:javascript",
+		"onLanguage:javascriptreact",
+		"onCommand:eslint.executeAutofix"
 	],
 	"main": "./out/extension",
 	"contributes": {
@@ -33,6 +35,11 @@
 					"type": "boolean",
 					"default": true,
 					"description": "Control whether eslint is enabled for JavaScript files or not."
+				},
+				"eslint.enableAutofixOnSave": {
+					"type": "boolean",
+					"default": false,
+					"description": "Control the flag to enable executing Autofix when you save a file."
 				},
 				"eslint.options": {
 					"type": "object",
@@ -49,6 +56,13 @@
 			{
 				"fileMatch": ".eslintrc.json",
 				"url": "http://json.schemastore.org/eslintrc"
+			}
+		],
+		"commands": [
+			{
+				"command": "eslint.executeAutofix",
+				"title": "Execute Autofix",
+				"category": "ESLint"
 			}
 		]
 	},
--- a/eslint/package.json
+++ b/eslint/package.json
@@ -39,7 +39,7 @@
 				"eslint.enableAutofixOnSave": {
 					"type": "boolean",
 					"default": false,
-					"description": "Control the flag to enable executing Autofix when you save a file."
+					"description": "Control whether eslint errors are auto-fixed when a file is saved."
 				},
 				"eslint.options": {
 					"type": "object",
--- a/eslint-server/src/lib/eslint.ts
+++ b/eslint-server/src/lib/eslint.ts
@@ -1,0 +1,33 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+export interface ESLintAutofixEdit {
+	range: [number, number];
+	text: string;
+}
+
+export interface ESLintProblem {
+	line: number;
+	column: number;
+	severity: number;
+	ruleId: string;
+	message: string;
+	fix?: ESLintAutofixEdit;
+}
+
+export interface ESLintDocumentReport {
+	filePath: string;
+	errorCount: number;
+	warningCount: number;
+	messages: ESLintProblem[];
+	output?: string;
+}
+
+export interface ESLintReport {
+	errorCount: number;
+	warningCount: number;
+	results: ESLintDocumentReport[];
+}
--- a/eslint-server/src/lib/protocol.ts
+++ b/eslint-server/src/lib/protocol.ts
@@ -25,12 +25,6 @@ export interface ESLintAutofixParams {
 	uri: string;
 
 	/**
-	 * The flag to indicate that this request target is being hidden.
-	 * If this is `true`, the server would rewrite the target file directly.
-	 */
-	hidden: boolean;
-
-	/**
 	 * The flag to indicate this request was triggered by saved.
 	 * If this is `true` and the `eslint.enableAutofixOnSave` configure is `false`, the request would be ignored.
 	 */
--- a/eslint-server/src/lib/protocol.ts
+++ b/eslint-server/src/lib/protocol.ts
@@ -1,0 +1,49 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import {RequestType} from 'vscode-languageserver';
+
+/**
+ * The autofix method is sent from the client to the server.
+ */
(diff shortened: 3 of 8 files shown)


mysticatea:
What should I do?

anyong:
@dbaeumer any chance we can get this PR merged?

kumarharsh:
+1 this would be a great addition!

juanfranblanco:
Since nobody is doing anything about this, can you not release your own version?

egamma:
The vscode-eslint extension is maintained by the vscode team and we were totally busy to get the march milestone done. Now it is and the plan is that the extensions we own will get more love from us next early next week after the easter holiday. 

Therefore please hold back with publishing a separate eslint extensions until we had the time to look at this PR.

dbaeumer:
@mysticatea first I apologize for this taking so long. Second, thanks a lot for the pull request. I like the way how it is coded specially the fact that you found out to send custom requests to the server.  

Regarding the code and I have a couple of comments:

- to avoid any save conflicts (e.g that the server saves the file and vscode does) we should avoid any save actions from the server. Otherwise we could end up in a situation where VSCode tries to save and the save is rejected because the files timestamp is different and VSCode then shows a diff editor

- replacing the whole contents of a file isn't nice for markers. For example a breakpoint set in the file is 'removed' when the whole content is replaces. It is better to apply small edits since it allows to adjust and keep these markers. To my knowledgr ESLint for each problem that is auto fixables reports a text edit for that problem. Need to investigate if this can be used.

- Since the server runs async we also need to ensure the edits we receive from the server can still be applied to the text document (e.g the user could have already changed the document again). This can be done by checking the TextDocument#version id when the edit is created and comparing it with the version id of the document before the server is invoked.

@mysticatea please let me know if you are willing to look into this. I have full understanding if you say no. The PR 'laid' around for a long time.

mysticatea:
Thank you for the review!
I will update this PR within this weekend.

eslint/package.json
```diff
@@ -34,6 +36,11 @@
 					"default": true,
 					"description": "Control whether eslint is enabled for JavaScript files or not."
 				},
+				"eslint.enableAutofixOnSave": {
+					"type": "boolean",
+					"default": false,
+					"description": "Control the flag to enable executing Autofix when you save a file."
```
kumarharsh: Maybe this can be reworded as "Control whether eslint errors are auto-fixed when a file is saved." or something similar?

mysticatea:
At first, I rebased this PR.

> to avoid any save conflicts (e.g that the server saves the file and vscode does) we should avoid any save actions from the server. Otherwise we could end up in a situation where VSCode tries to save and the save is rejected because the files timestamp is different and VSCode then shows a diff editor

In my understanding, document objects and editor objects are not 1 vs 1. Non-displayed documents do not have their editor object, then I cannot edit those documents from extensions.
The reason to rewrite files from the server was that I wanted to support auto-fixing when "Save All" command was executed as well. I had thought it's no problem if the editor was nothing and if the timing is immediately after the document was saved.

Anyway, I removed this.
Is there a way to apply auto-fixing to hidden documents instead?

> replacing the whole contents of a file isn't nice for markers. For example a breakpoint set in the file is 'removed' when the whole content is replaces. It is better to apply small edits since it allows to adjust and keep these markers. To my knowledgr ESLint for each problem that is auto fixables reports a text edit for that problem. Need to investigate if this can be used.

I got it. I have not noticed breaking points are lost on edit.

Surely, each problem has the fixing content.
I rewrote the fixing logic with the same way as eslint.

I could not find API to modify Unicode BOM, so it's remained as TODO.

> Since the server runs async we also need to ensure the edits we receive from the server can still be applied to the text document (e.g the user could have already changed the document again). This can be done by checking the TextDocument#version id when the edit is created and comparing it with the version id of the document before the server is invoked.

That's what I wanted to know.
I updated with using `version`.

@kumarharsh Thank you for the correction! I updated the text.

dbaeumer:
@mysticatea Thanks a lot for the updated PR. I will have a look today & tomorrow. There is one thing I would like to explore and sorry for not mentioning this earlier but it didn't cross my mind when doing the first review: we do have the concept of code actions which fixes code and are bound to diagnostics (a little bit comparable to what eslint does). However it currently doesn't support fix on save and it doesn't support bulk apply. But I would like to bring these things together since it will help not only the eslint extension. I will add my findings to this PR.

dbaeumer:
The other advantage for using code action is that we would get the light bulb under errors / warnings for free.

dbaeumer:
This is  how it would look like with code actions and light bulbs
capture

capture

mysticatea:
@dbaeumer I know code actions and I have an implementation of using code actions for ESLint auto-fixing: https://github.com/eslint/eslint/issues/5040#issuecomment-174387295
I think I will send a PR after this in my mind. May I add it to this PR?

Applying code actions on saved makes sense more.
But I have a concern, ESLint auto-fixing cannot apply all items sometimes. But it may be hard detecting whether it can apply or not on code actions, I think.

dbaeumer:
A proper implementation of this requires https://github.com/Microsoft/vscode/issues/239 which we try to tackle for April/May.

The problem is that we can't say that the next save we listen to is the one triggered by the autofix since the workbench has auto save support which can easily conflict with this assumption.

dbaeumer:
@mysticatea: I code up a auto fix support that is totally based on code actions. Although it required some work on the language client / server first I think it is the right direction to go. I created a branch with the code: https://github.com/Microsoft/vscode-eslint/tree/dbaeumer/autofix-support

I would love to get your feedback on this. We mainly took the approach of code actions to implement a story that other linter implementers can copy from. The advantage of the code action approach is from our standpoint as follows:

- one UI gesture to apply a single fix, all fixes of a special kind of all fixes to a file (see screen shot below). We can even think about a generic command that could be reused across all language to trigger all code actions for a whole file or even workspace.

- with Microsoft/vscode#239 in reach we can even implement a 'generic' apply fixes on save. All we needed to do is to provide the proper code actions.

capture

mysticatea:
@dbaeumer Great!

- I think the 3 commands on code actions make sense.
- This implement seems to be smarter than the current `eslint --fix`. This can fix some cases that `eslint --fix` cannot fix. But if there is one overlapped problem, this implement failed to fix all fixable problems silently. On the otherhand, `eslint --fix` would fix all problems other than overlapped problems.
- If there are problems of some kind at the same location, we may need to consider the display:
  image

dbaeumer:
@mysticatea do you have an easy example with overlapping problems. I am pretty sure I can do something about it.

dbaeumer:
@mysticatea got it. No need to provide and example.

dbaeumer:
Added code to sort ot overlapping edits

mysticatea:
Perfect!
Thank you so much, @dbaeumer .

mysticatea:
Oops!

I had modified `package.json` for `Execute Autofix` command and `eslint.enableAutofixOnSave` option but now it's not working.

dbaeumer:
@mysticatea this was my fault. I left them in but they don't work. For the auto fix on save to work correctly we need https://github.com/Microsoft/vscode/issues/239 which is either planned for April or May. Otherwise we can cause harm since the whole eventing and save is async.

For the action I opened: https://github.com/Microsoft/vscode/issues/5650. This ensures that we don't get a apply quick fix action per language / linter

OysteinAmundsen:
Hi. 

It's been a couple of months now. Any chance this will be released soon? It will be a really useful feature, and I'm really looking forward to it. 

:-)

dbaeumer:
@OysteinAmundsen the feature is available since a while. I simply forgot to close the PR. Look here for how it looks: https://github.com/Microsoft/vscode-eslint/pull/19#issuecomment-208471376

OysteinAmundsen:
@dbaeumer so... no autofix on save?

dbaeumer:
@OysteinAmundsen unfortunately not. It still depends on https://github.com/Microsoft/vscode/issues/239.

PR #38, opened 2016-03-01
Chrisdias/license

chrisdias:
The new "LICENSE" button in the command palette will open the [license in the marketplace](https://marketplace.visualstudio.com/items/dbaeumer.vscode-eslint/license):

image

In order for the marketplace to scrape the license you need to set the license property of package.json to "SEE LICENSE IN <<< filename >>>". This PR does this.

### DIFF
--- a/eslint/package.json
+++ b/eslint/package.json
@@ -4,7 +4,7 @@
 	"description": "Integrates ESLint into VS Code.",
 	"version": "0.10.12",
 	"author": "Microsoft Corporation",
-	"license": "MIT",
+	"license": "SEE LICENSE in License.txt",
 	"repository": {
 		"type": "git",
 		"url": "https://github.com/Microsoft/vscode-eslint.git"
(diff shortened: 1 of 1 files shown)


Tyriar:
Won't this break then break the license detection in npm?

Can a custom `licenseFile` field in package.json be used for this instead?

chrisdias:
It shouldn't, this is standard according to the [package.json](https://docs.npmjs.com/files/package.json) docs.

CC @joaomoreno

Tyriar:
Well it will probably work but it's technically not following the spec since it's MIT which has an SPDX identifier. Could we not use the SPDX identifier in the `ext install`? That would be much easier than clicking through to a web page.

chrisdias:
If you use the SPDX identifier you will get a generic license, correct? I know Joao did some work here (see https://github.com/Microsoft/vscode-vsce/issues/66).

In this case we want the slightly customized MIT license found in License.txt, which "SEE LICENSE IN License.txt" should pick up.

chrisdias:
Talked with @joaomoreno. With https://github.com/Microsoft/vscode-vsce/issues/66 `vsce` will now find the custom License.txt, so this PR is no longer necessary and we just need to republish the extension. I will open a new issue on that.

PR #41, opened 2016-03-02
Use Settings instead of Configuration

scharf:
Explain where to change the `configuration`, which actually is a `setting`.


egamma:
Hi michael @scharf :smile: 

The wording makes it more accurate. One issue I've see is that users are putting eslint rules into the settings instead of into an eslintrc config file. I therefore suggest to also improve the wording for

From:
>`eslint.options`: an option bag as defined by the ESLint API [here](http://eslint.org/docs/developer-guide/nodejs-api#cliengine). Defaults to an empty option bag. 

To: 
>`eslint.options`: options to configure how eslint is started using the ESLint [CLI Engine API](http://eslint.org/docs/developer-guide/nodejs-api#cliengine). Defaults to an empty option bag.

scharf:
Hi Erich,

I updated the PR with your suggestion.

egamma:
:+1:

dbaeumer:
Thanks @scharf :-)

PR #90, opened 2016-06-17
Changed the description for indexing

egamma:
Added a reference to JavaScript so that the extension is found when looking for JS extensions.


dbaeumer:
@egamma this only changes the readme. Something missing ?

egamma:
vsce is smarter than I thought and the change to the README is not needed - closing.

PR #115, opened 2016-08-24
Support for specifying eslint module path

### LINKED ISSUE #87: Make eslint library location configurable
I'm excited to use this extension. I'm on `VSCode@1.2.0` and `vscode-eslint@0.10.16`.

I have a somewhat unusual setup where `eslint` is brought in by a larger static analysis package that I've built. My dep tree looks like this:

```
root-package
└─┬ static-analysis-tool
  └── eslint
```

When running from `root-package`, I get the following error message:

> Failed to load eslint library. Please install eslint in your workspace folder using 'npm install eslint' or globally using 'npm install -g eslint' and then press Retry.

I would like to solve this by adding an option in the user preferences file to point vscode-eslint to the correct eslint location. This would be similar to how I can use the `typescript.tsdk` option to tell VSCode which TypeScript package to use.

This would also be useful for using VSCode to develop eslint itself. In that case, you'd want to use the eslint in the cwd.

I'd be happy to send a PR for this if you're amenable to it.

Thanks!

Related:
* #32

kumarharsh:
This change adds a setting called `eslint.relativeModulePath`
which is helpful in the cases when the node_modules folder
might not reside in the root of the workspace.

Fixes #87

### DIFF
--- a/eslint-server/src/server.ts
+++ b/eslint-server/src/server.ts
@@ -151,13 +151,17 @@ documents.onDidClose((event) => {
 });
 
 connection.onInitialize((params): Thenable<InitializeResult | ResponseError<InitializeError>>  | InitializeResult | ResponseError<InitializeError> => {
-	let rootPath = params.rootPath;
-	let initOptions: { legacyModuleResolve: boolean; } = params.initializationOptions;
+	let initOptions: { legacyModuleResolve: boolean; relativeModulePath: string; } = params.initializationOptions;
 	let noCLIEngine = new ResponseError(99, 'The eslint library doesn\'t export a CLIEngine. You need at least eslint@1.0.0', { retry: false });
 	let result: InitializeResult = { capabilities: { textDocumentSync: documents.syncKind, codeActionProvider: true }};
 	let legacyModuleResolve = initOptions ? !!initOptions.legacyModuleResolve : false;
 	let resolve = legacyModuleResolve ? Files.resolveModule : Files.resolveModule2;
-
+	let rootPath;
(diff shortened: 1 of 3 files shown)


kumarharsh:
@dbaeumer this also works well with the new `resolveModule` function too, but I literally just saw that you'd pushed that change to master, and monkey testing hasn't revealed any bugs. Do both the resolveModules work the same way?

dbaeumer:
@kumarharsh thanks a lot for the PR. Unfortunately it crosses with the work I currently do. With the new resolve method I already look up the parent chain. So the normal use case for having the eslint installed in a parent folder already works.

I also added support for a nodePath value which I add to the NODE_PATH env value before resolving the module. So if the eslint is installed in a wired location people can no point to that. 

So I would opt to see have far we get with these improvements. I would only add a another relative path if the improvements are not enough.

Any objections to close the PR.

kumarharsh:
@dbaeumer that's nice! Just one question: what about the cases when the javascript project is nested within the root workspace folder, so eslint is buried deep in the folder tree. Do the changes you're doing take that case into account?

kumarharsh:
Hi @dbaeumer, it seems like the changes you've described are not yet published in the plugin, and I'm not able to run the `vsce package` command to get a vsix package to work with

dbaeumer:
@kumarharsh in that case you can use the new eslint.nodePath setting to point to the eslint library to use. 

Regarding publishing: the changes are available as 2.4.2-next.x packages for the language server / client. We are currently in endgame and I will publish a new version when testing and fixing is done.

kumarharsh:
OK, thank you :)

PR #146, opened 2016-09-26
preserve focus when showing output channel

bpasero:
I would like to propose that we do not pass focus to the output channel when we open it automatically.

@dbaeumer fyi


dbaeumer:
Thanks @bpasero. To my defense I have to say that it is strange that a panel that doesn't allow typing takes focus by default in the first place. Wouldn't it make more sense to never take focus on show and have a separate method focus.

bpasero:
@dbaeumer totally, now you have to convince @jrieken to behave like that and we are all set

PR #153, opened 2016-10-10
Fixes zero location support

mysticatea:
- `problem.line`, `problem.column`, `problem.endLine`, and
`problem.endColumn` may be zero.
- In that case, `vscode-eslint` seems to crash silently.
- This PR modifies the location calculation to avoid the crash.


dbaeumer:
Thanks for the PR.

Issue #183, opened 2016-12-08
Bad contrast of ES Lint in status bar


bpasero:
![image](https://cloud.githubusercontent.com/assets/900690/21017612/3321f460-bd6a-11e6-9604-54b732ec3954.png)

dbaeumer:
This got changes a while ago to a yellowish color.

PR #206, opened 2017-02-16
Update statusBarItem color

chenxsan:
When ESLint is not available, we should update the statusBarItem color to inform user. I think it will also fix https://github.com/Microsoft/vscode-eslint/issues/183.

screen shot 2017-02-16 at 9 48 56 am

### DIFF
--- a/eslint/extension.ts
+++ b/eslint/extension.ts
@@ -214,7 +214,7 @@ export function realActivate(context: ExtensionContext) {
 				statusBarItem.color = 'yellow';
 				break;
 			case Status.error:
-				statusBarItem.color = 'darkred';
+				statusBarItem.color = '#aaa';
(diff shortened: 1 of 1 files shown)


dbaeumer:
@chenxsan I see the need for an updated color here however making the error case gray is not a good idea (the icon would get gray for example if the .eslintrc file can not be parsed). I would still go for some red tone.

chenxsan:
@dbaeumer I tried a lot of other colors, most of them looks bad when using with a blue background color :), with grey we could tell user the function is kind of disabled.

Anyway, if you come up with a better one, I would love to copy to [vscode-standardjs](https://github.com/chenxsan/vscode-standardjs/blob/master/eslint/extension.ts#L166).

PR #228, opened 2017-04-21
Fix resolving eslint when NODE_PATH has multiple entries

silverlyra:
For my workspace setup I need to specify two paths in `NODE_PATH`. This should work, except currently the call to `Files.resolve` also uses `NODE_PATH` for the resolution process's current working directory, and cd'ing to `/somewhere:/somewhere/else` fails.

It looks like both `directory` or `workspacePath` are reasonable cwd choices instead; I picked `directory` to match the Promise rejection case below.

### DIFF
--- a/eslint-server/src/server.ts
+++ b/eslint-server/src/server.ts
@@ -351,7 +351,7 @@ documents.onDidOpen((event) => {
 			let file = uri.fsPath;
 			let directory = path.dirname(file);
 			if (nodePath) {
-				 promise = Files.resolve('eslint', nodePath, nodePath, trace).then<string>(undefined, () => {
+				 promise = Files.resolve('eslint', nodePath, directory, trace).then<string>(undefined, () => {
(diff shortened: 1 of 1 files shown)


silverlyra:
👉  *poke*

silverlyra:
@dbaeumer pretty please?

dbaeumer:
Sorry, but I am busy shipping VS Code May release. Will look into this beginning of next week.

dbaeumer:
Actually this can't be fixed that way. Due to the resolution rules of node passing anything else than nodePath as the cwd might pick up modules from a different location than nodePath. Node resolve always starts in the current working directory for node modules and then walks up the parent chain. And the contract of nodePath is that it is preferred over any other location.

IMO the fix for this has to go into the vscode-languageserver-node repository which provides Files.resolve. Are you willing to look into this?

silverlyra:
actually, in the meantime I stopped using VSCode. :woman_shrugging: I'll just close this out.

Issue #291, opened 2017-09-09
Activating extension `dbaeumer.vscode-eslint` failed: vscode_languageclient_1.ProposedProtocol is not a function.


chenxsan:
I'm trying to run the source code on my computer, only got this:

<img width="1915" alt="screen shot 2017-09-09 at 5 20 02 pm" src="https://user-images.githubusercontent.com/1091472/30238901-3a84420c-9583-11e7-945f-3eb83ccc115b.png">

And I checked repo https://github.com/Microsoft/vscode-languageserver-node/tree/master/client, there is no `ProposedProtocol` exported.

1. Node.js - 8.4.0
2. npm - 5.4.1
3. VSCode - 1.16.0

Any idea?

dbaeumer:
We had some last minute changes in the LSP API that the ESLint extension needed to adopt.

PR #292, opened 2017-09-09
Fix compile errors

chenxsan:
This pull request should fix https://github.com/Microsoft/vscode-eslint/issues/291, there's still one compile error:

screen shot 2017-09-09 at 9 49 33 pm

I could remove it from the `import` and define an interface to fix it:

```ts
interface WorkspaceFolder {
	name: string;
	uri: string;
}
```
But I'm not sure it's the right way, so I'll leave it.

All of the fixes are collected from https://github.com/Microsoft/vscode-extension-samples, hope it helps.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -15,8 +15,8 @@ import {
 	TextDocumentIdentifier, NotificationType, ErrorHandler,
 	ErrorAction, CloseAction, State as ClientState,
 	RevealOutputChannelOn, VersionedTextDocumentIdentifier, ExecuteCommandRequest, ExecuteCommandParams,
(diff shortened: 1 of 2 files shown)


chenxsan:
Oh I just checked the build history here https://travis-ci.org/Microsoft/vscode-eslint/builds/273215705, seems everything is fine with the old code. I'm not sure what's going on on my test now. Btw, My Node.js version is 8.4.0, and VSCode 1.16.0. Any suggest?

chenxsan:
Unfortunately, I tested it on Node.js 6.11.3, some compile errors as 8.4.0.

dbaeumer:
@chenxsan thanks for the PR. I fixed the server and the client and pushed the change.

Issue #208, opened 2017-02-20
Can I only auto-fix all error and do not auto-fix warnings ?


benpigchu:
Some warnings is auto-fixable but we do not want to fix it, like "prefer-const" .
Can we just fix errors when we auto-fix all ?

dbaeumer:
@benpigchu do you know if this is possible on the command line when calling eslint --fix?

benpigchu:
@dbaeumer NO, but it seems that we can do it with the eslint node.js api.

dbaeumer:
Cool. Interested in working on a PR to support this ?

medikoo:
Yes, in general it'll be great to be able to filter some rules specifically from auto-fixing.

It was discussed also at ESlint, and it's the recommended approach (solve it on editor side).
See the [thread](https://github.com/eslint/eslint/issues/7549#issuecomment-314041266)

dbaeumer:
@medikoo thanks for the pointer.

AdenFlorian:
Would something similar to this work here? https://github.com/Microsoft/vscode-tslint/pull/231

dbaeumer:
@AdenFlorian having `autoFixOnSave` to be a string[] alternatively and then only autofix the onces listed makes sense to me.

TasukuUno:
Hi, It is better for me to specify rules that **disable** auto-fix like [linter-eslint (atom)](https://github.com/AtomLinter/linter-eslint/blob/5ed08f1c655729b383abf76b6347e585b2055c3d/package.json#L101-L108)! I think it is hard to set all of the other fixable rules...

dbaeumer:
@AdenFlorian @TasukuUno comment makes sense to me. May be we can use the `!` to negate strings like in git config files.

AdenFlorian:
@dbaeumer What should happen if you have this config?
```json
autoFixOnSave: [
  "semi",
  "!quotes"
]
```

Mixing includes and excludes complicates things a bit, so I made a table to help define the behavior of whatever we decide. Populated with my current implementation. I'll come and think about this more later.

| autoFixOnSave | behavior |
| ---------------- | ---------- |
| false                 | never auto fix on save |
| true                  | auto fix everything on save |
| array                | auto fix only rules in the array on save |

davidvalachovic:
@dbaeumer @TasukuUno @benpigchu  Thoughts?

dbaeumer:
[...]

PR #325, opened 2017-10-28
autoFixOnSave may now be array of specific rules to fix

AdenFlorian:
Somewhat implements #208


herbrandson:
Any progress on this? We'd really love this feature

AdenFlorian:
@herbrandson We're still trying to come up with a solution that gets everyone what they want (see #208). We could use some ideas there. Once we get a solid plan, I'll be good to implement it.

dbaeumer:
I am closing the PR due to inactivity.

PR #332, opened 2017-11-01
Remove unused import (TextEditor)

phi2039:
The most recent commit breaks the build (via a linter rule) by introducing an unused import. This PR removes that import.


dbaeumer:
Got already fixed.

PR #407, opened 2018-02-20
Update README.md to feature a tasks 2.0.0 example

jackrl:


dbaeumer:
@jackrl thanks for the PR. A better 2.0.0 task definition would be to have a named task and not use a single global task. That allows giving it a nice name. Something like

```
		{
			"label": "ESLint",
			"type": "shell",
			"windows": {
				"command": ".\\node_modules\\.bin\\eslint *.js"
			},
			"linux": {
				"command": "./node_modules/.bin/eslint *.js"
			},
			"osx": {
				"command": "./node_modules/.bin/eslint *.js"
			},
			"presentation": {
				"reveal": "silent",
			},
			"problemMatcher": "$eslint-stylish"
		}

```

Even better would be if the ESLint extension would be a task provider and provides this task out of the box without the need of having an entry in the tasks.json file. You could use this as an example on how to do this is you are interested: https://github.com/Microsoft/vscode-extension-samples/blob/master/task-provider-sample\src\extension.ts#L14

dbaeumer:
Would you be willing on working on such a task provider.

jackrl:
@dbaeumer I created something that should work: https://github.com/jackrl/vscode-eslint/tree/workspace-lint-task

I added a new configuration, 'eslint.enableLintTask', with a false default so that the people who don't need or want it are not affected.

It is a shame that you can't open the 'problems' tab after the task is done, or at least I wasn't able to find how it would be done.

Let me know if you want a separate PR for it. I suppose that if this is added there is no need for the task example in the README,md anymore.

dbaeumer:
There is a issue in VS Code to allow tasks to open the problems pane when errors are detected. So no need to file a new one.

dbaeumer:
@jackrl can you create a PR for the work in https://github.com/jackrl/vscode-eslint/tree/workspace-lint-task

jackrl:
Done #410

dbaeumer:
I will close this PR then. I removed the tasks.json from the README.

PR #410, opened 2018-02-22
Create a task provider for linting the whole workspace

jackrl:


jackrl:
Would tere be a need for multipme tasks that could be made based on a configuration object?

I think being able to customize at least  the arguments of the command would be nice.

Which on should I try implementing if any?

dbaeumer:
Very cool. Thanks a lot

dbaeumer:
@egamma  FYI

dbaeumer:
@jackrl I did some tweaks to the code. The task provide is called whenever the Run Task command is executed. So there is not need to do anything special when the configuration changes other than creating the task and setting it to undefined.

jackrl:
@dbaeumer Did your tweaks get lost somewhere? I don't see them on the master branch yet.

By the way, what about the possibility of configuring the task or creating multiple tasks base in config? Would that be a good feature in your opinion? I could work on that at some point.

dbaeumer:
@jackrl haven't checked them in yet since I noticed that you only supported a folder workspace and not a multi folder workspace. Needed to do more tweaks. Will check in soon.

dbaeumer:
Checked in the changes. I moved the code to a new file tasks.ts since to support multi folder more code was necessary. Let me know what you think.

jackrl:
It is much cleaner in its own file and creating the tasks when running the run task command is convenient too. I'll check it in more detail in a bit.

jackrl:
@dbaeumer Now it's great. Just what what the previous manually added task did but easier and always ready.

PR #423, opened 2018-03-12
Use eslint --init to generate configuration

### LINKED ISSUE #409: Add Command to create a YAML configuration
The extension currently has a command to `Create '.eslintrc.json' file`, but I would prefer to use `.eslintrc.yml`.

Is it possible to add a command that generates a YAML configuration with the same content?

Hirse:
The extension currently has a command to `Create '.eslintrc.json' file`, but I would prefer to use `.eslintrc.yml`.
This PR adds functionality to generate a Yaml file with the same settings as the JSON.

closes #409

### DIFF
--- a/client/src/command.ts
+++ b/client/src/command.ts
@@ -1,0 +1,21 @@
+import * as fs from 'fs';
+import * as path from 'path';
+
+function exists(file: string): Promise<boolean> {
+	return new Promise<boolean>((resolve, _reject) => {
+		fs.exists(file, (value) => {
+			resolve(value);
+		});
+	});
+}
+
+export async function findEslint(rootPath: string): Promise<string> {
+	const platform = process.platform;
+	if (platform === 'win32' && await exists(path.join(rootPath, 'node_modules', '.bin', 'eslint.cmd'))) {
+		return path.join('.', 'node_modules', '.bin', 'eslint.cmd');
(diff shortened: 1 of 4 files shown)


dbaeumer:
@Hirse thanks for providing this. However I am not a big fan of it to provide a VS Code command to create a YAML file. I would rather let the ESLint extension contribute a task to run eslint --init in the terminal which at the end will as you to pick the format you want to use. This is more flexible.

Would you be willing to look into this?

Hirse:
@dbaeumer I agree that using `eslint --init` is cleaner. 
Please see the last commit where I have replaced the config generation with running that command in the terminal.

dbaeumer:
@Hirse thanks. 

Some comments:

- to make this work in a multi folder setup the folder location must be passed as the cwd to the terminal. Otherwise the configuration might be created in the wrong location.
- what happens if eslint is only installed locally should the command then be `.\node_modules\.bin\eslint` instead.

Hirse:
@dbaeumer Thanks for your comments and sorry for the delay.

I am passing the cwd to the terminal and reusing the existing logic to find the location of the ESLint location. I have extracted that functionality to a new file as it doesn't entirely belong to the tasks.

Hirse:
Thanks for merging, but I just saw that the Readme needs to be updated:
https://github.com/Microsoft/vscode-eslint/blob/4303aa4075905ffc2c37dc4bee7e36d4526953ad/README.md#L7

I might be able to open a PR later today if you didn't get to it by then.

dbaeumer:
Done.

PR #466, opened 2018-05-10
Add ESLint document formatter and document range formatter

### LINKED ISSUE #417: "Format Document" command should use ESLint
There didn't seem to be any mention of this functionality, except a by implication in "formatOnSave" issues.

VSCode has a JavaScript formatter, but ESLint should replace it, not use a secondary one with a different conflicting command.

Maxwellewxam:
Fixes #417 .

The client-side code is inspired from prettier-vscode.
The server-side code introduces 2 new commands for retrieving the TextEdits. The ranged command isn't very elegant as it iterates over all edits and filters those in range, but if I understand well the codeActions behavior it loads all the edits of the current file and stores them anyway.
I'm new to VSCode extension creation and typescript so tell me if I do things the wrong way.


dbaeumer:
@Maxwellewxam thanks for the PR. Since the ESLint server is a full LSP server in terms of the request it supports (see https://github.com/Microsoft/language-server-protocol/) the server can directly implement the `formatting`and `rangeFormatting`request and return the corresponding edits. To do so do the following:

- have a setting as you do. In a first cut you can send the setting in the initialize request. We will see later how we can make this more dynamic.
- in the server initialized notification register a `formatting` and `rangeFormatting` depending on that setting.
- implement the corresponding handler by registering them on the server side. Something like this:
```
messageQueue.registerRequest(DocumentFormattingRequest.type, (params) => {
	let result: TextEdit[] = [];
	...
	return result;
});
```

Ping me if something is unclear.

dbaeumer:
@Maxwellewxam quick feedback. There is no need to implement a `ESLintEditProvider` on the client side. If the server registers these providers the VS Code `LanguageClient` will do the request automatically for you.

dbaeumer:
The easiest to get it working is to register them as a result of the initialize request. Return something like:
```
connection.onInitialize((_params) => {
	return {
		capabilities: {
			textDocumentSync: {
				openClose: true,
				change: TextDocumentSyncKind.Full,
				willSaveWaitUntil: true,
				save: {
					includeText: false
				}
			},
			codeActionProvider: true,
			executeCommandProvider: {
				commands: [CommandIds.applySingleFix, CommandIds.applySameFixes, CommandIds.applyAllFixes, CommandIds.applyAutoFix]
			},
			documentFormattingProvider: true,
			documentRangeFormattingProvider: true
		}
	};
});
```

However this would ignore a potential setting. But we can work on this later.

server/src/eslintServer.ts
```diff
@@ -1142,5 +1145,33 @@ messageQueue.registerRequest(ExecuteCommandRequest.type, (params) => {
...
+	let textDocument: VersionedTextDocumentIdentifier = {
+		uri: params.textDocument.uri,
+		version: documents.get(params.textDocument.uri).version
+	};
+	let edits = computeAllFixes(textDocument);
+	if (!edits) {
+		return [];
+	}
+	let rangedEdits = edits.filter((edit) => {
+		if (params.range.start.line < edit.range.start.line && edit.range.start.line < params.range.end.line) {
```
Coder-256: Is this line correct? Consider this instead:
```suggestion
		if (params.range.start.line <= edit.range.start.line && edit.range.end.line <= params.range.end.line) {
```

dbaeumer:
@Coder-256 thanks for doing the review.

ValYouW:
@dbaeumer hi, any plans for this PR or range formatting should already work (its not working for me... :( ) thx

dbaeumer:
What is implemented in the meantime is formatting. It must be enabled via a setting. I looked into range formatting as well but that is not supported by ESLint itself so I decided to stay way from it. 

I will close the PR.

Issue #459, opened 2018-04-25
Add syntax highlighting for .eslintignore file


garyking:
Please add syntax highlighting for .eslintignore files, mostly so comments are highlighted.

I don't know if it should just use the `Shell script` format, or the same one as used in .gitignore files, etc.?

dbaeumer:
Actually VS Code has no syntax coloring for .gitignore. Or do you see that colored ?

garyking:
I'm using the [gitignore](https://marketplace.visualstudio.com/items?itemName=codezombiech.gitignore) extension for gitignore files, so perhaps the eslint ext could do the same and color eslintignore files the same way?

Hirse:
@dbaeumer, can we close this one now?

dbaeumer:
Yes, we can. thanks for pinging.

PR #473, opened 2018-05-24
eslintignore comment syntax highlighting

Hirse:
Colors comments and enables commenting of lines using keyboard shortcut.

see #459


dbaeumer:
@aeschli can you please have a quick review. You know a lot more about this than I do.

aeschli:
The change looks good!

@Hirse If you wanted you could make a similar addition for `gitignore` files as we don't have any colorization there as well.
We could share the grammar of the two, if we're sure that the syntaxes are identical.

Hirse:
@aeschli Do you mean in vscode? 
If we add that, would it make sense to define some kind of generic `ignore-file` language for `gitignore`, `eslintignore`, `npmignore`, and possibly others?

aeschli:
@Hirse A generic ignore-file language makes sense. We could still place it along with the git extension

dbaeumer:
That makes sense to me as well. That would mean for eslint we would then simply reference the genernal syntax.

Hirse:
@aeschli do you mean adding a definition in [vscode `extensions/git/languages`](https://github.com/Microsoft/vscode/tree/master/extensions/git/languages) or adding a new `ignore` extension which only contains this language definition?

aeschli:
@Hirse I'd suggest to add it to vscode/extensions/git

dbaeumer:
@aeschli do you recommend that we wait with this in ESLint and then reuse the one that goes into git?

Hirse:
@dbaeumer Sorry for the delay, I have changed the code to reference the existing `ignore` language as @aeschli did in https://github.com/Microsoft/vscode/commit/eb2b51aeb9cb76d1a1ff24b015c38e5bcb90cc95 so this should be good to go now.

dbaeumer:
Thanks!.

PR #516, opened 2018-08-10
Allow setting the node runtime to use for the language server. Fixes #345

### LINKED ISSUE #345: “Unexpected token ...” (object spread) when using `.eslintrc.js`
Hello. This eslint config is valid JavaScript but makes the ESLint server crash with `Unexpected token ...`.
```javascript
const myGlobals = { globalA: 'A', globalB: 'B' }

module.exports = {
  globals: {
    LIVE: true,
    PRODUCTION: true,
    DEVELOPMENT: true,
    ...myGlobals
  }
}
```

I guess it isn’t using my current node version (latest, v9) but some other version that doesn’t support spreading? When I start the VSCode debugger, it says it is indeed using v9… I’m not too sure what’s happening as I expect my extensions to use my node version?

I could obviously use `Object.assign` instead, but… 😄

Mossop:
Per the discussion in #345 it is useful toi be able to select the node version to run ESLint as. This adds the eslint.server.runtime setting which can be set to the path of the node binary to use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ This extension contributes the following variables to the [settings](https://cod
 ```
 - `eslint.run` - run the linter `onSave` or `onType`, default is `onType`.
 - `eslint.autoFixOnSave` - enables auto fix on save. Please note auto fix on save is only available if VS Code's `files.autoSave` is either `off`, `onFocusChange` or `onWindowChange`. It will not work with `afterDelay`.
-- `eslint.server.runtime` - use this setting to set the path of the node runtime to run ESLint under.
+- `eslint.runtime` - use this setting to set the path of the node runtime to run ESLint under.
 - `eslint.nodePath` - use this setting if an installed ESLint package can't be detected, for example `/myGlobalNodePackages/node_modules`.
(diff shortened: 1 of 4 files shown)


Mossop:
I'm not really sure what the travis failures are all about, they seem unrelated to my changes :(

dbaeumer:
@Mossop nice work. One minor request. Can we name the setting `eslint.runtime` (without server). There are other server settings that don't have `server` in its name (like `eslint.nodePath`).

README.md
```diff
@@ -22,6 +22,7 @@ This extension contributes the following variables to the [settings](https://cod
 ```
 - `eslint.run` - run the linter `onSave` or `onType`, default is `onType`.
 - `eslint.autoFixOnSave` - enables auto fix on save. Please note auto fix on save is only available if VS Code's `files.autoSave` is either `off`, `onFocusChange` or `onWindowChange`. It will not work with `afterDelay`.
+- `eslint.server.runtime` - use this setting to set the path of the node runtime to run ESLint under.
```
dbaeumer: I suggest to name this `eslint.runtime`. `eslint.nodePath` is for the server as well and we never put server into the naming so far.

Issue #245, opened 2017-05-25
Idea: quick fix to disable rule for selected line(s)


OliverJAsh:
I make frequent use of comments to disable selected ESLint rules for selected lines, for example:

``` js
// eslint-disable-next-line max-len
{} // long line of JavaScript …
```

It would be great if VSCode provided a "quick fix" option in the UI to silence rules it is complaining about by automatically adding these comments for the selected lines. I get quite tired of writing them, especially looking up the precise rule names.

Has this idea come up before?

dbaeumer:
Would be a good opportunity for a PR

wmertens:
Yes, I often wish for this too. There could be a button on the tooltip "disable for this line".

I'd love to help but don't know where to start. Basically there should be two commands, `disable-eslint-linebefore` and `disable-eslint-sameline`, and they should insert or amend the currently hovered or cursored error into the editor.

fiznool:
Would love to see this too. The same feature is currently available in WebStorm:

<img width="501" alt="screen shot 2018-01-11 at 10 07 19" src="https://user-images.githubusercontent.com/820863/34819994-4155360a-f6b7-11e7-9256-3cddabfcbbe2.png">

agrande:
Is anybody working on this?
I switched over from WebStorm to VSCode recently, and I miss that feature very much.

dbaeumer:
Not that I am aware of. I can point you to the right code locations in case you want to give it a try.

agrande:
Yes please, I might do just that, if it’s not too complex (I’m new to VSCode, I’d have to read up a bit about developing extensions).
The way I see it, that’d be a couple of additional entries in the Quickfix command contextual menu to suppress the error/warning for the line or the whole file (or even the eslint config file).

dbaeumer:
@agrande that is correct, however the fixes need to be contribute in the ESLint linting server code. We run the linters outside of VS Code itself in its own process. The corresponding code is here: https://github.com/Microsoft/vscode-eslint/blob/master/server/src/eslintServer.ts#L1015

So instead of only adding the fixes we would need to:

- have commands for fixes that add an ignore comment.
- add code action results for these commands.

bendtherules:
Just a temp solution I use -
Add eslint disable snippets as user snippet in VSCode

```json
{
	"Disable eslint rule next line": {
		"prefix": "eslint-disable-next",
		"body": [
			"eslint-disable-next-line ${1:<rule-name>}"
		],
[...]

PR #530, opened 2018-08-25
Added Suppress rule for this line / Suppress rule for file quick actions / Show documentation

loune:
* Added Suppress rule for this line / Suppress rule for file quick actions (#245)

eslint

* Added show documentation quick action

* Fix possible issue with incorrect command run when 2 rule single fixes are available. This is due to `CommandIds.applySingleFix` getting overriden inside `commands` by the last added single fix quick action. The fix is to include the `ruleId` in the key when adding to `commands`.

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -1086,18 +1086,20 @@ messageQueue.registerRequest(CodeActionRequest.type, (params) => {
 		let indentationText = /^([ \t]*)/.exec(lineText)[1];
 		workspaceChange.getTextEditChange({uri, version: editInfo.documentVersion}).add(createDisableLineTextEdit(editInfo, indentationText));
 		commands.set(`${CommandIds.applyDisableLine}:${disableRuleId}`, workspaceChange);
+		let title = `Suppress ${disableRuleId} for this line`;
 		result.push(CodeAction.create(
-			`Suppress ${disableRuleId} for this line`,
-			Command.create(editInfo.label, CommandIds.applyDisableLine, disableRuleId),
+			title,
+			Command.create(title, CommandIds.applyDisableLine, disableRuleId),
 			CodeActionKind.QuickFix
 		));
 
 		workspaceChange = new WorkspaceChange();
 		workspaceChange.getTextEditChange({uri, version: editInfo.documentVersion}).add(createDisableFileTextEdit(editInfo));
 		commands.set(`${CommandIds.applyDisableFile}:${disableRuleId}`, workspaceChange);
+		title = `Suppress ${disableRuleId} for the entire file`;
 		result.push(CodeAction.create(
-			`Suppress ${disableRuleId} for the entire file`,
-			Command.create(editInfo.label, CommandIds.applyDisableFile, disableRuleId),
+			title,
+			Command.create(title, CommandIds.applyDisableFile, disableRuleId),
 			CodeActionKind.QuickFix
 		));
 	};
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -194,7 +194,7 @@ function makeDiagnostic(problem: ESLintProblem): Diagnostic {
 	};
 }
 
-interface AutoFix {
+interface FixableProblem {
 	label: string;
 	documentVersion: number;
 	ruleId: string;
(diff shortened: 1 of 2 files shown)


loune:
@dbaeumer any chance this could be merged? Thanks

server/src/eslintServer.ts
```diff
@@ -799,6 +827,14 @@ function validate(document: TextDocument, settings: TextDocumentSettings, publis
 		if (publishDiagnostics) {
 			connection.sendDiagnostics({ uri, diagnostics });
 		}
+
+		// cache documentation urls for all rules
+		ruleDocUrls.clear();
```
dbaeumer: Is it necessary to cache this on every validate which is actually run very frequently or could we cache this once on only clear the cache if the .eslintrc file changes?

server/src/eslintServer.ts
```diff
@@ -673,7 +691,15 @@ connection.onInitialize((_params) => {
...
 			executeCommandProvider: {
-				commands: [CommandIds.applySingleFix, CommandIds.applySameFixes, CommandIds.applyAllFixes, CommandIds.applyAutoFix]
+				commands: [
+					CommandIds.applySingleFix,
+					CommandIds.applySameFixes,
+					CommandIds.applyAllFixes,
+					CommandIds.applyAutoFix,
+					CommandIds.applyDisableLine,
+					CommandIds.applyDisableFile,
+					CommandIds.openRuleDoc,
```
dbaeumer: This is not necessary since the is sent from the server to the client and not the other way.

dbaeumer: I mean CommandIds.openRuleDoc

server/src/eslintServer.ts
```diff
@@ -1045,15 +1093,49 @@ messageQueue.registerRequest(CodeActionRequest.type, (params) => {
...
+				editInfo.label,
+				Command.create(editInfo.label, CommandIds.applySingleFix, ruleId),
+				CodeActionKind.QuickFix
+			));
+		}
+
 		let workspaceChange = new WorkspaceChange();
-		workspaceChange.getTextEditChange({uri, version: documentVersion}).add(createTextEdit(editInfo));
-		commands.set(CommandIds.applySingleFix, workspaceChange);
+		let lineText = textDocument.getText(Range.create(Position.create(editInfo.line - 1, 0), Position.create(editInfo.line - 1, 255)));
```
dbaeumer: You can use Integer.MAX_VALUE here instead of 255 to be on the save side. Will have the same effect.

server/src/eslintServer.ts
```diff
@@ -181,7 +198,8 @@ interface AutoFix {
 	label: string;
```
dbaeumer: Since with the change we now store problem in general per file shouldn't we rename this. Something that contains problem in the name

dbaeumer: FixableProblem might be an idea.

server/src/eslintServer.ts
dbaeumer: Makes sense. I misunderstood what you wanted to achieve

dbaeumer:
Very nice work.

loune:
Thanks @dbaeumer !

dbaeumer:
I did some minor tweaks. One thing I forgot about is that different files can have different .eslintrc files (in different workspace folders) so we need to remember the rules per file. Changed this and pushed.

Will publish a new version of ESLint beginning of next week.

Friksel:
Great improvement! Thanks a lot for this!

TimShilov:
Is there an option to hide those additional menu items?
I usually don't suppress rules at all and I don't need the Documentation link. When I open Light Bulb menu I want to only see recommendations and for me those new menu items are just noise.

dbaeumer:
Actually that is a fair request. @loune would you be willing to add such a setting ?

Friksel:
@dbaeumer hereby: https://github.com/Microsoft/vscode-eslint/issues/573

loune:
Sure, I can have a look at adding a setting.

alexiosd:
Hello, great addition, but what happened to Fix All of a specific problem?
The Fix all auto-fixable problems could be problematic if you don't know what it will fix.

thank you.

loune:
@alexiosd this patch shouldn't change how Fix All of a specific problem works. I'll see what's wrong with that as well.

PR #562, opened 2018-10-21
Add link to ESLint rules via url attribute of the Diagnostic API

### LINKED ISSUE #151: Add a link to ES Lint rule page.
image

thibaudcolas:
Fixes #151. Relies on https://github.com/Microsoft/vscode/pull/61436 to implement diagnostic URLs in VS Code (fixing https://github.com/Microsoft/vscode/issues/11847), as well as https://github.com/Microsoft/vscode-languageserver-node/pull/432.

This PR does three things:

1. Add `eslint-rule-documentation` to retrieve URLs as a fallback to the rules metadata, so older versions of ESLint / older versions of plugin rules can also benefit from URLs.
2. Make the URL retrieval work with versions of ESLint before v4.15.0 (2018-01-06), which introduced  `CLIEngine#getRules()`.
3. Use the (new) `url` attribute of diagnostics to forward the ESLint rule's URL to VS Code, for display wherever.

Changes 1. and 2. could be useful now without further changes in VS Code or anywhere else. I can make separate PRs for these if they seem desirable enough.

Here is what the end result looks like at the moment (WIP):
wip-highlight-over-link

---

The instructions to try this out locally are at the end of https://github.com/Microsoft/vscode/pull/61436.

### DIFF
--- a/server/src/eslint-rule-documentation.d.ts
+++ b/server/src/eslint-rule-documentation.d.ts
@@ -1,0 +1,7 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+declare module "eslint-rule-documentation" {
+	export default function ruleURI(ruleId: string): { found: boolean, url: string };
+}
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -17,6 +17,7 @@ import {
 } from 'vscode-languageserver';
 
 import URI from 'vscode-uri';
+import * as ruleURI from 'eslint-rule-documentation';
 import * as path from 'path';
 import {EOL} from 'os';
 
@@ -851,10 +852,17 @@ function validate(document: TextDocument, settings: TextDocumentSettings, publis
 		// cache documentation urls for all rules
 		if (!ruleDocData.handled.has(uri)) {
 			ruleDocData.handled.add(uri);
-			cli.getRules().forEach((rule, key) => {
+			cli.getRules().forEach((rule, ruleId) => {
+				let url;
+
 				if (rule.meta && rule.meta.docs && Is.string(rule.meta.docs.url)) {
-					ruleDocData.urls.set(key, rule.meta.docs.url);
+					url = rule.meta.docs.url;
+				} else {
+					// @ts-ignore TODO: Should type-check the 3rd-party module.
+					url = ruleURI(ruleId).url;
 				}
+
+				ruleDocData.urls.set(ruleId, url);
 			});
 		}
 	} finally {
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -202,8 +202,15 @@ function makeDiagnostic(ruleURLs: Map<string, string>, problem: ESLintProblem): 
 	};
 
 	if (problem.ruleId) {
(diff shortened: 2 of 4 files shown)


server/src/eslint-rule-documentation.d.ts
```diff
@@ -0,0 +1,7 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+declare module "eslint-rule-documentation" {
+	export default function ruleURI(ruleId: string): { found: boolean, url: string };
```
thibaudcolas: This isn't the correct type signature – the function is a CommonJS `module.exports`. I couldn't figure out how to type those correctly.

server/src/eslintServer.ts
```diff
@@ -199,6 +200,20 @@ function makeDiagnostic(problem: ESLintProblem): Diagnostic {
...
+	if (problem.ruleId) {
+		let url = ruleURLs.get(problem.ruleId);
+
+		if (!url) {
+			// @ts-ignore TODO: Should type-check the 3rd-party module.
+			url = ruleURI(problem.ruleId).url;
+			ruleURLs.set(problem.ruleId, url);
+		}
+		// @ts-ignore TODO Ignoring "property URL does not exist" for now. It will exist in the newest version of vscode.
+		diagnostic.url = url;
```
thibaudcolas: This type error will go away when relying on a version of VS Code that has a URL attribute for diagnostics. For local devs, I was using an `npm link`ed version, but that didn't seem to be enough.

server/src/eslintServer.ts
```diff
@@ -819,6 +834,26 @@ function validate(document: TextDocument, settings: TextDocumentSettings, publis
 		}
 
 		let cli = new settings.library.CLIEngine(newOptions);
+
+		// cache documentation urls for all rules
+		if (cli.getRules) {
```
thibaudcolas: As far as I can tell, the implementation in master would fail for ESLint versions below v4.15.0 (jan 2018).

server/src/eslintServer.ts
```diff
@@ -199,6 +200,20 @@ function makeDiagnostic(problem: ESLintProblem): Diagnostic {
...
 		code: problem.ruleId
 	};
+
+	if (problem.ruleId) {
+		let url = ruleURLs.get(problem.ruleId);
+
+		if (!url) {
+			// @ts-ignore TODO: Should type-check the 3rd-party module.
+			url = ruleURI(problem.ruleId).url;
+			ruleURLs.set(problem.ruleId, url);
```
thibaudcolas: If the URL isn't in the cache, it's not from the rule's metadata. So we use eslint-rule-documentation directly, and then set its result in the cache.

server/src/eslint-rule-documentation.d.ts
dbaeumer: A good doc on how to create these declaration files can be found here: https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html

thibaudcolas: Arr thank you, I had looked at these docs but totally missed the relevant template (https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-function-d-ts.html). Will update this later.

thibaudcolas:
Closing because of https://github.com/Microsoft/vscode/pull/61436#issuecomment-437942788.

PR #565, opened 2018-10-27
Add option keepErrorCodes to keep rule ids

vikr01:
In d98bfd60fcbb1be6112bcd6b1e9a4fc9eab318f9, the rule id causing `eslint` errors was removed from displaying.

That's *the* most valuable feature to me, so I've made an option `eslint.keepErrorCodes` to re-enable it. It expects a boolean, and defaults to `false`.


vikr01:
@dbaeumer At Google, the recommended VSCode settings includes `"update.channel": "none"` (disables auto updates). So for legacy reasons, an option like this would be useful for people using a slightly older version of VSCode.

dbaeumer:
@vikr01 The latest ESLint extension should only install on 1.28 upwards (which has the code in the problems view, but not in the hover). Older versions of VS Code will still use an older version of ESLint which still adds the text to the problem message.

Is this tolerable for Google. If if we add the switch the publish ESLint extension will only install on 1.28 forward

vikr01:
@dbaeumer That should be good. Thanks!

PR #567, opened 2018-10-30
If eslint not listed as dependency use global one only

### LINKED ISSUE #549: Don't use local eslint unless it's explicitly listed in package.json
If I leave `eslint.nodePath` set to `null` (the default), and I edit an NPM project that does not have `eslint` explicitly listed in its `package.json`, yet one of its deps includes `eslint`, then this VSC extension will still use the local eslint rather than the global one. This is not preferred, because then global plugins will not be loaded.

janowsiany:
Fixes #549

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -413,7 +413,7 @@ function resolveSettings(document: TextDocument): Thenable<TextDocumentSettings>
 					return Files.resolve('eslint', settings.resolvedGlobalPackageManagerPath, directory, trace);
 				});
 			} else {
-				promise = Files.resolve('eslint', settings.resolvedGlobalPackageManagerPath, directory, trace);
+				promise = Files.resolve('eslint', settings.resolvedGlobalPackageManagerPath, undefined, trace);
 			}
 		} else {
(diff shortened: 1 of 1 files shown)


janowsiany:
ping @dbaeumer

server/src/eslintServer.ts
```diff
@@ -413,7 +413,7 @@ function resolveSettings(document: TextDocument): Thenable<TextDocumentSettings>
 					return Files.resolve('eslint', settings.resolvedGlobalPackageManagerPath, directory, trace);
 				});
 			} else {
-				promise = Files.resolve('eslint', settings.resolvedGlobalPackageManagerPath, directory, trace);
+				promise = Files.resolve('eslint', settings.resolvedGlobalPackageManagerPath, undefined, trace);
```
dbaeumer: This only results in the fact the local installed eslint version is never considered even if it is not listed in the package.json.

dbaeumer:
See also https://github.com/Microsoft/vscode-eslint/issues/549#issuecomment-440717379

dbaeumer:
I will close the PR since it has no traction right now. Please ping if you want to continue on this.

PR #569, opened 2018-10-31
Improve pipeline

angulito:
Improve the pipeline using jobs instead of a single script to build all the pipeline commands.

Also I've splitted the `npm run compile` to run the `compile:client` and `compile:server` so it's easier to find if there is any compile problem.

See pipeline example -> https://travis-ci.org/angulito/vscode-eslint/builds/448863504

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ sudo: false
 node_js:
   - "6.11.2"
 
-install:
-  - npm install
-
-script:
-  - npm run compile
+jobs:
+  include:
+    - stage: "compile"
+      name: Compile
+      script: npm run compile
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ node_js:
 
 jobs:
   include:
-    - stage: "compile"
-      name: Compile
-      script: npm run compile
+    - stage: "compile client"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,8 +4,12 @@ sudo: false
 node_js:
   - "6.11.2"
 
-install:
-  - npm install
```
dbaeumer: How do we now make sure npm install is executed. I have to admit that I am not very familiar with pipelines :-)

angulito: You can see an example [here](https://travis-ci.org/angulito/vscode-eslint/jobs/448863505)
image

The fact is if you have `jobs` in the pipeline, they are related to the language - it means travis ci knows what are the default steps for the javascript/nodejs language:
to install: `npm install`
to run the tests: `npm tests`
and for each job, the pipeline always runs the default command for the install step, in this case, `npm install`.

For that reason you can see [here](https://travis-ci.org/angulito/vscode-eslint/builds/448863504) that for the `compile client` and `compile server` jobs, travisCI runs the npm install command inside (and you cannot see that command declared in the pipeline)

dbaeumer: Thanks for the explanation !

PR #572, opened 2018-11-04
Issue 529, Added support for pnpm

Lee182:


server/src/eslintServer.ts
```diff
@@ -355,31 +356,36 @@ let connection = createConnection();
...
-function globalYarnPath(): string {
-	if (_globalYarnPath === void 0) {
-		_globalYarnPath = Files.resolveGlobalYarnPath(trace);
-		if (_globalYarnPath === void 0) {
-			_globalYarnPath = null;
+	},
+	npm: {
+		cache: undefined,
+		get(): string {
+			const npmPath = execSync('npm root -g').toString().trim();
```
dbaeumer: Can we keep using Files.resolveGlobalNodePath here to make sure we are not breaking anything.

server/src/eslintServer.ts
```diff
@@ -355,31 +356,36 @@ let connection = createConnection();
...
+			return npmPath
+		}
+	},
+	pnpm: {
+		cache: undefined,
+		get(): string {
+			const pnpmPath = execSync('pnpm root -g').toString().trim();
```
dbaeumer: Using the global package manager path often result in unexpected behavior (for example the path being different than expected). This is why both methods on Files take a trace value which when on traces the value. We whould do the same for pnpm. However having the method here makes sense. The corresponding trace code looks like this:

```ts
	try {
		process.on('SIGPIPE', handler);
		let stdout = spawnSync(npmCommand, ['config', 'get', 'prefix'], options).stdout;

		if (!stdout) {
			if (tracer) {
				tracer(`'npm config get prefix' didn't return a value.`);
			}
			return undefined;
		}
		let prefix = stdout.trim();
		if (tracer) {
			tracer(`'npm config get prefix' value is: ${prefix}`);
		}

		if (prefix.length > 0) {
			if (isWindows()) {
				return path.join(prefix, 'node_modules');
			} else {
				return path.join(prefix, 'lib', 'node_modules');
			}
		}
		return undefined;

```

server/src/eslintServer.ts
Lee182: Yeah that's fine :+1:

server/src/eslintServer.ts
Lee182: yeah looks good to me. I wasn't sure to how you handle errors for spawn commands, so left that blank, I figured you'd have an more official way of logging spawned errors. I'm willing to test the branch when it's updated

client/src/extension.ts
```diff
@@ -617,31 +616,30 @@ export function realActivate(context: ExtensionContext) {
 			let state = context.globalState.get<NoESLintState>(key, {});
 			let uri: Uri = Uri.parse(params.source.uri);
 			let workspaceFolder = Workspace.getWorkspaceFolder(uri);
-			let packageManager = Workspace.getConfiguration('eslint', uri).get('packageManager', 'npm');
+			const packageManager = Workspace.getConfiguration('eslint', uri).get('packageManager', 'npm');
+			const localInstall = {
+				npm: 'npm install eslint',
+				pnpm: 'pnpm install pnpm',
```
TomasHubelbauer: Shouldn't this say `pnpm install eslint`?

dbaeumer: Correct. Will fix it.

ruslanguns: @dbaeumer this PR has been merged but this change wasn't added

dbaeumer: What exactly are you referring to?

adpirz: Was fixed here: https://github.com/microsoft/vscode-eslint/commit/545149c03f90824d5f5a8a6bfdcb13e4249d77a3

PR #619, opened 2019-02-09
Allow relative node runtime.

antoniodejesusochoasolano:
Setting `'eslint.runtime'` to a relative path wasn't working. VS Code ESLint would end up trying to run the relative path from `/` - e.g. if `'eslint.runtime': './node_modules/.bin/node'`, VS Code ESLint would try running `/./node_modules/.bin/node`.


dbaeumer:
Actually this fix is very misleading since the 'eslint.runtime' is currently defined as a window setting and not as a workspace setting. Making it relative to a workspace folder as proposed is not predictable. 

We can't make the setting a resource setting since it would require to start different servers if different runtimes are configured. Adding such a support will be very costly.

Not sure yet what a good solution for this is. May be a window setting that lets a user define which folder to look into for relative paths. Something like

`"eslint.runtime": "${folderName}/node_modules/.bin/node"`

PR #665, opened 2019-04-24
Update README.md to include bit about multi-project workspaces

aleccool213:
I know this is fairly specific but it gave me trouble. Feel free to tell me this is the wrong spot to put this.


dbaeumer:
@aleccool213 before adding this I would like to understand what the actually issue is and whether this might be a problem of the extension itself. Could you open a normal issue with steps how to reproduce (best a GitHub repository to clone). Then I investigate. Please also notice that in a multi workspace folder each folder must have its own self contained eslint configuration.

aleccool213:
@dbaeumer I agree that this piece of doc I want to add is not giving the complete picture. I will create a repo reproducing the issue I had and explain it it's README.md the solution.

It could very well be an eslint issue and not this extension but the error was raised by the extension so it's hard to determine where the appropriate place for a doc add is needed.

Issue #682, opened 2019-06-03
Lint task should honor eslint.options setting


dbaeumer:
For the details see: https://github.com/microsoft/vscode-eslint/issues/91#issuecomment-497033867

PR #698, opened 2019-06-27
fix https://github.com/microsoft/vscode-eslint/issues/682

salvofid:
**eslint.lintTaskOptions** added: this options allows to specify a string that is used as argument list during the task linting the whole workspace

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -77,6 +77,7 @@ interface TextDocumentSettings {
 	autoFixOnSave: boolean;
 	quiet: boolean;
 	options: any | undefined;
+	lintTaskOptions: string;
 	run: RunValues;
 	nodePath: string | undefined;
 	workspaceFolder: WorkspaceFolder | undefined;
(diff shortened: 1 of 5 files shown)


client/src/extension.ts
```diff
@@ -77,6 +77,7 @@ interface TextDocumentSettings {
 	autoFixOnSave: boolean;
 	quiet: boolean;
 	options: any | undefined;
+	lintTaskOptions: string;
```
dbaeumer: This is not necessary since the linter task is executed on the client side only. So the settings don't to travel to the server.

dbaeumer:
@salvofid thanks for the PR. See my comments above.

client/src/extension.ts
salvofid: Right, parameter removed :-). Thanks

sw-double:
LGTM?

dbaeumer:
I did some minor additional tweaks. Named the option eslint.lintTask.options and renamed eslint.provideLintTask to eslint.lintTask.enable.

If options are provided no `.` is added to the command line. It must be part of options.

PR #753, opened 2019-09-12
Fix None position values in publishDiagnostics message

### LINKED ISSUE #752: language server sends invalid position sometimes in publishDiagnostics
For an issue like `Newline required at end of file but not found`, eslint doesn't report endLine or endChar and current code is wrong in checking those properties for null but not for undefined. So server ends up sending `null`/`None` value for `line` and `character` properties in `Position` object which is invalid according to the spec [1] (only numbers are allowed).

[1] https://microsoft.github.io/language-server-protocol/specification#position

rchl:
For some issues eslint triggers error with no `endLine` and/or
`endColumn` defined. In that case the code should use `line` and
`column` for the values, respectively. While the code has attempted to
do that, it only checked for `null` value for those while eslint
actually doesn't even have them set so they are undefined.

Fix by checking that value is either null or undefined.

This problem causes random exceptions on some LSP clients (for me it
triggered in Sublime LSP). Not sure if it affected Vetur.

Resolves #752

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -29,6 +29,10 @@ namespace Is {
 		return value === true || value === false;
 	}
 
+	export function nullOrUndefined(value: any): value is boolean {
+		return value === null || value === undefined;
+	}
+
 	export function string(value: any): value is string {
 		return toString.call(value) === '[object String]';
 	}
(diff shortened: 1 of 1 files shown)


dbaeumer:
Thanks!

PR #764, opened 2019-09-25
Add `typescriptreact` to the README's TypeScript section

kourge:
VS Code wasn't linting my `.tsx` files even though running the CLI `eslint` with the correct glob patterns did the job. It took me thirty minutes to realize that this is because `tsx` files fell under the purview of `typescriptreact`, and so that also needed to be in `eslint.validate` for things to work as expected.


dbaeumer:
Makes sense.

PR #766, opened 2019-09-30
Register ESLint as a document formatter

### LINKED ISSUE #154: 'Fix all auto-fixable problems' doesn't fix as many issues as possible
I have ESLint extension 1.0.7 with VSCode 1.6.0 in win10 x64
and ESLint 3.7.1 globally installed.

First of all, please note that since [ESLint 2.9.0](http://eslint.org/blog/2016/04/eslint-v2.9.0-released#improved-autofix): 

> when you use the --fix option, ESLint will make multiple passes over your code in an attempt to fix as many issues as possible.

<br>

Let's say I have the following .eslintrc.js in `C:\Users\Kostas`:

```
module.exports = {
    "env": {
        "browser": true,
        "es6": true,
        "greasemonkey": true,
        "jquery": true
    },
    "extends": "eslint:recommended",
    "rules": {
        "indent": ["error","tab"],       
    }
};
```

Now, I have this Javascript file:

```
var localTimezone;

    function convertDates(f) {
        if (f === 'f1') {
            format = 'DD.MM.YYYY HH:mm';
        } else {
            format = 'MMMM DD, YYYY';
        }
    }
```

If I do `eslint --fix myfile.js` via CLI  (a single time) it will be fixed  into:

```
var localTimezone;

function convertDates(f) {
    if (f === 'f1') {
        format = 'DD.MM.YYYY HH:mm';
    } else {
        format = 'MMMM DD, YYYY';
    }
}

```

But, in VSCode the ``Eslint: Fix all auto-fixable problems`  command doesn't fix all indentation at once:
executing it the first time will only fix indentation on line 3, 
a second time fixes lines 4 + 9, 
and finally a 3rd time fixes  lines 5-8, 
i.e. it requires multiple (3) executions.

_For reference, in Atom with linter 1.11.18 and linter-eslint 8.0.0 using my global ESLint installation as well,
with the relevant command (`Linter eslint: Fix File`) all indentation is fixed with a single execution._

### LINKED ISSUE #417: "Format Document" command should use ESLint

dtinth:
This pull request makes `vscode-eslint` register itself as a document formatter (fixes #417). When the **Format Document** is invoked, it will fix all possible issues using the same API as `eslint --fix` (fixes #154).

Note that the original “Fix all auto-fixable problems” has been left intact with the original behavior as described in #154.

For more information, see this comment that analyzes the differences between registering a command vs a document formatter and how it relates to #154: https://github.com/microsoft/vscode-eslint/issues/154#issuecomment-533626122

### Screenshots

Using “Fix all auto-fixable problems”

before

Using “Format document” (implemented in this PR)

after

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -13,7 +13,7 @@ import {
 	Command, WorkspaceChange,
 	CodeActionRequest, VersionedTextDocumentIdentifier,
 	ExecuteCommandRequest, DidChangeWatchedFilesNotification, DidChangeConfigurationNotification,
-	WorkspaceFolder, DidChangeWorkspaceFoldersNotification, CodeAction, CodeActionKind, Position
+	WorkspaceFolder, DidChangeWorkspaceFoldersNotification, CodeAction, CodeActionKind, Position, DocumentFormattingRequest
 } from 'vscode-languageserver';
 
 import { URI } from 'vscode-uri';
@@ -179,6 +179,7 @@ interface ESLintReport {
 
 interface CLIOptions {
 	cwd?: string;
+	fix?: boolean;
 	fixTypes?: string[];
 }
 
@@ -250,6 +251,7 @@ function computeKey(diagnostic: Diagnostic): string {
 }
 
 let codeActions: Map<string, Map<string, FixableProblem>> = new Map<string, Map<string, FixableProblem>>();
+let fixedFileOutput: Map<string, string> = new Map();
 function recordCodeAction(document: TextDocument, diagnostic: Diagnostic, problem: ESLintProblem): void {
 	if (!problem.ruleId) {
 		return;
@@ -757,7 +759,8 @@ connection.onInitialize((params, _cancel, progress) => {
 					CommandIds.applyDisableFile,
 					CommandIds.openRuleDoc,
 				]
-			}
+			},
+			documentFormattingProvider: true
 		}
 	};
 });
@@ -782,7 +785,7 @@ const singleErrorHandlers: ((error: any, document: TextDocument, library: ESLint
 	showErrorMessage
 ];
 
-function validateSingle(document: TextDocument, publishDiagnostics: boolean = true): Thenable<void> {
+function validateSingle(document: TextDocument, publishDiagnostics: boolean = true, runFixes: boolean = false): Thenable<void> {
 	// We validate document in a queue but open / close documents directly. So we need to deal with the
 	// fact that a document might be gone from the server.
 	if (!documents.get(document.uri)) {
@@ -793,7 +796,7 @@ function validateSingle(document: TextDocument, publishDiagnostics: boolean = tr
 			return;
 		}
 		try {
-			validate(document, settings, publishDiagnostics);
+			validate(document, settings, publishDiagnostics, runFixes);
 			connection.sendNotification(StatusNotification.type, { state: Status.ok });
 		} catch (err) {
 			let status = undefined;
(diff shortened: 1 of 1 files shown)


dbaeumer:
@dtinth thanks for the PR. The format document should as the auto fix does provide a minimal set of edits and not simply replace the whole file content. Replacing the whole content has in most editors bad side effects like loosing break points and causing unnecessary re rendering of stuff. So it should reuse `computeAllFixes`.

dtinth:
@dbaeumer Thanks for the comment.

> Replacing the whole content has in most editors bad side effects like loosing break points and causing unnecessary re rendering of stuff.

Have you verified your assumption? [Based on my testing and analysis](https://github.com/microsoft/vscode-eslint/issues/154#issuecomment-533626122), we _can_ replace the whole content because according to VS Code’s source code, it will take the full document edit [and compute a more minimal edits for us](https://github.com/microsoft/vscode/blob/66dc7a11b709c9353d6bb2fd144c7e088ad35ecf/src/vs/editor/common/services/editorSimpleWorker.ts#L445), so we don’t have to. This only happens when we register it as a document formatter, not a command, and that’s why I created this PR. 😉

This is also [how extensions like prettier-vscode](https://github.com/prettier/prettier-vscode/blob/894edf36c6e52b9b1f6137514c0c46b7d13e14dc/src/PrettierEditProvider.ts#L296) and [vscode-go](https://github.com/microsoft/vscode-go/blob/cd30b0dcd05318223c472294623e5e5777a28810/src/goFormat.ts#L85-L89) work. Breakpoints stay intact.

If we use `computeAllFixes`, we end up with a problem in #154, where we have to run the command multiple times before the fix converges.

dbaeumer:
@dtinth yes, that is true for VS Code (and I know it since my team mate wrote the code :-)) but I am not a fan on relying on this. I do see the problems around #154 but the right approach here is to address this once in the server. Otherwise we end up with different fixing behavior whether things get fixed on format or on save. It would also mean that when running format on save we always send the whole document back. I see that this is an easy fix but I still think addressing the problem of running the command multiple times is the right direction for the extension in general.

What we could do is to let eslint do the looping by passing in the fix option and then let the server compute the diff. Might be easier than shifting positions and merging edits.

dtinth:
@dbaeumer I see, thanks for the insight. It would be great if VS Code’s diffing function is exposed to extensions or to the language server protocol library… then implementing this would be much easier.

> yes, that is true for VS Code

Hm… is this extension intended to be used by other editors as well? (I get a glimpse that this may be the case when you said “Replacing the whole content has _in most editors_ bad side effects” in an earlier comment.) I thought the name was `vscode-eslint` so I didn’t concern myself with other editors at all.

> Otherwise we end up with different fixing behavior whether things get fixed on format or on save.

What we can also do is deprecate `eslint.autoFixOnSave` and `eslint.executeAutofix` commands entirely, like [how `vscode-go` deprecated `go.formatOnSave` two years ago](https://github.com/microsoft/vscode-go/commit/e95eed681077442a42979f91402d642118798daf).

Anyways, I think I have accomplished my goal for this PR, which is to provide a workaround for a long-standing issue. For end users who wants the issue #154 fixed: You can clone [the source repository](https://github.com/microsoft/vscode-eslint), apply [my patch](https://github.com/microsoft/vscode-eslint/pull/766.patch), then build and install the extension to your own VS Code installation. The instructions to build the project is [here](https://github.com/microsoft/vscode-eslint/blob/master/contributing.md) and you can generate a VSIX file to install on your local VS Code by following the [instructions here](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#packaging-extensions).

Since I now have other priorities, I have no intention to change the approach in this PR. Maybe someone else in the community can help continue this work 😃. For maintainers: Please feel free to close this PR, merge it, or leave it open 😉.

remychvn:
We really need a solution to fix #154
This implementation **may not** be the best one, but it seems to correct a problem that has been going on for years!
Many of us use "prettier" to bypass this problem. This complicates the setup of a new project, the "prettier" rules can conflict with the "eslint" rules...

dbaeumer:
Sorry for the late reply. I was out on vacation. I will look into borrowing some code from VS Code to create a minimal diff so that we can fix this for formatting and quick fixes in one go.

thernstig:
@dbaeumer Any progress? :)

dbaeumer:
Yes, I pushed some code yesterday. No new ESLint version yet but will come soon.

https://github.com/Microsoft/vscode-eslint/blob/dbaeumer/formatting/server/src/eslintServer.ts#L1545

I will close the PR since the new code computes a minimal diff.

PR #797, opened 2019-11-06
Implements fix filters

### LINKED ISSUE #208: Can I only auto-fix all error and do not auto-fix warnings ?

arcanis:
This diff fixes #208 the "right" way. Instead of relying on a third-party plugin set (which isn't very practical, especially when you add third-party plugins such as `react` or `import` into the mix), we simply filter out the lint results that don't pass the predicate set in `options.fix` (similar to what the CLIEngine already allows).

Instead of doing an eval as suggested in another issue, I went with a simpler approach where each field can list the values it accepts. This first iteration doesn't include blacklisting as I felt like it would be overkill and very simple to implement later if we think we need it.

Usage:
```json
{
    "eslint.options": {
        "fix": {
            "severity": 2,
            "ruleId": "prettier/prettier"
        }
    }
}
```

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -183,6 +183,7 @@ interface ESLintReport {
 
 interface CLIOptions {
 	cwd?: string;
+	fix?: boolean | {[key: string]: any};
 	fixTypes?: string[];
 }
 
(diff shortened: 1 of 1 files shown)


dbaeumer:
@arcanis actually this conflicts with the new implementation of fix all on save which now calls ESLint to do this due to the n iteration concept in ESLint. So for the fix all on save I think we should allow users to point to a different eslintrc file to have the full flexibility to configure this. For the code action in the light bulb menu I am not sure if it is actually need to filter them. What is your take?

haomaigege:
> This diff fixes #208 the "right" way. Instead of relying on a third-party plugin set (which isn't very practical, especially when you add third-party plugins such as `react` or `import` into the mix), we simply filter out the lint results that don't pass the predicate set in `options.fix` (similar to what the CLIEngine already allows).
> 
> Instead of doing an eval as suggested in another issue, I went with a simpler approach where each field can list the values it accepts. This first iteration doesn't include blacklisting as I felt like it would be overkill and very simple to implement later if we think we need it.
> 
> Usage:
> 
> ```json
> {
>     "eslint.options": {
>         "fix": {
>             "severity": 2,
>             "ruleId": "prettier/prettier"
>         }
>     }
> }
> ``

> @arcanis actually this conflicts with the new implementation of fix all on save which now calls ESLint to do this due to the n iteration concept in ESLint. So for the fix all on save I think we should allow users to point to a different eslintrc file to have the full flexibility to configure this. For the code action in the light bulb menu I am not sure if it is actually need to filter them. What is your take?

@arcanis can you solve the conflicts, this feature seems to be great

PR #798, opened 2019-11-07
Allow specifying a set of rules in autoFixOnSave

### LINKED ISSUE #767: Allow specifying a subset of rules to be autofixed on save
With `eslint.autoFixOnSave` on, ESLint applies all the fixes to a file while I only need just few of them to be applied automatically and have ability to decide whether the rest is ignored or fixed via Quick Fix.

It might be achieved via supporting `boolean | string[]` as a value for `autoFixOnSave`, as it was done in older version of [TSLint plugin](https://github.com/microsoft/vscode-tslint/blob/1f23aa3ca1e14120db3bd7675532df26327927b1/tslint-server/src/tslintServer.ts#L28).

I'd be happy to send a PR if you consider this change useful

pablobirukov:
`autoFixOnSave` supports `boolean` and `string[]` values.

`boolean` behaves as before, with a `string[]` value set auto fix is on but the plugin will only apply fixes from rules listed as a values of `autoFixOnSave` field.

Fixes #767

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -74,7 +74,7 @@ interface TextDocumentSettings {
 	validate: boolean;
 	packageManager: 'npm' | 'yarn' | 'pnpm';
 	autoFix: boolean;
-	autoFixOnSave: boolean;
+	autoFixOnSave: boolean | string[];
 	quiet: boolean;
 	options: any | undefined;
 	run: RunValues;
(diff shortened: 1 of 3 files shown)


Sebastianokunkelo:
evolution-gaming:autoFixOnSave-specific-rules

dbaeumer:
This got addressed with a setting for codeActionsOnSave.

PR #845, opened 2019-12-18
Add reload command.

### LINKED ISSUE #164: command to restart server or plugins without quitting vscode
This may be a rare use case, but I'm actually *developing* an eslint plugin. I live test the plugin by symlinking the package to node_modules, and load it up in my eslintrc file. So I develop and live test at the same time.

When I make changes to the implementation of a rule for my plugin, it doesn't get picked up by this extension. Is there a way to restart the eslint server or tell it to reload the plugins?

I guess the ability to restart would also be useful if we update an eslint plugin's version with npm (to load up new rules, for example)

threehams:
Add reload command to restart the client without having to restart VS Code.

Fixes #164.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -1221,6 +1221,11 @@ function realActivate(context: ExtensionContext): void {
 			});
 		}),
 		Commands.registerCommand('eslint.showOutputChannel', () => { client.outputChannel.show(); }),
+		Commands.registerCommand('eslint.reload', () => {
+			client.stop().then(() => {
+				client.start();
+			});
+		}),
 		statusBarItem
 	);
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "vscode-eslint",
-	"version": "2.0.2",
+	"version": "2.0.7",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
--- a/package.json
+++ b/package.json
@@ -357,6 +357,11 @@
 				"title": "Show Output Channel",
 				"category": "ESLint",
 				"command": "eslint.showOutputChannel"
+			},
+			{
+			  "title": "Reload ESLint",
+			  "category": "ESLint",
+			  "command": "eslint.reload"
 			}
 		],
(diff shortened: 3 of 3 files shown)


Davilink:
I saw that you named it "Reload", it more like a restart than a "Reload". Normally restart is associated with the need of the service to shutdown then starting versus a reload that keep the service running but refresh the configuration.

MLefebvreICO:
Can we have this merge please ?

The number of times I switch branches and ESLint will throw warnings/errors on a  `typescript-eslint` rule is getting frustrating .. 

If at least I could : 

1. `F1`
2. `Reload Typescript server`
3. `F1`
4. `Reload EsLint server`
5. Continue working

But now I have to reload my window or restart VSCode.

The fact that this happens when I switch branch is irritating, but at least if we could get some sort of reload for ESLint, that would be great !

client/src/extension.ts
```diff
@@ -1221,6 +1221,11 @@ function realActivate(context: ExtensionContext): void {
 			});
 		}),
 		Commands.registerCommand('eslint.showOutputChannel', () => { client.outputChannel.show(); }),
+		Commands.registerCommand('eslint.reload', () => {
+			client.stop().then(() => {
+				client.start();
```
numfin, line 1226: async () => {
await client.stop();
await client.start();
}

rdperottoni:
Would love to have this feature.

threehams:
I'll publish a fork some time in the next week. I've needed this at work enough to justify spending the time on it there.

dbaeumer:
I looked into this today and this is not as easy as calling stop & start. Stop removes a lot of state from the client which makes the start not be successful in all cases. The right fix is actually to create a new client and not reuse the old instance. Need to think what needs to be done to make this successful.

Sebastianokunkelo (approved):

Sebastianokunkelo:
client/src/extension.ts

Sebastianokunkelo:
Sebastianokunkelo:master

dbaeumer:
@threehams I delivered this for 2.1.17 in a slightly different form.

PR #853, opened 2019-12-19
Make it clear that modes are arrays and not objects for config

NikhilVerma:
This change makes it clear that modes config is a part of an array and not an object on it's own. It took me a good while to figure out why eslint mode:auto wasn't working. Until I wrapped it in an array.


dbaeumer:
@NikhilVerma thanks!

PR #857, opened 2019-12-22
Coffeescript-specific comment syntax for disabling comments

### LINKED ISSUE #825: improve out of the box support for coffeescript
Hi all, 

there's now on `eslint`'s (and `prettier`'s btw) ecosystem side proper support for **coffeescript** per @helixbass 's **[eslint-plugin-coffee](https://github.com/helixbass/eslint-plugin-coffee)**  and **[prettier-plugin-coffeescript](https://github.com/helixbass/prettier-plugin-coffeescript)** ongoing (great) work.

This is all good and things work out of the box mostly on `vscode-eslint` side almost fully except in one case - when an issue pops, when quickfixing it *if* one wants to disable affected line or next one (calls to [`eslint.{createDisable{,Same}LineTextEdit}`](https://github.com/microsoft/vscode-eslint/blob/master/server/src/eslintServer.ts#L1456-L1462)) the inserted comment (with hints to `eslint`) is inserted with the wrong (from coffeescript's point of view) syntax, since `vscode-eslint` has   (on  https://github.com/microsoft/vscode-eslint/blob/master/server/src/eslintServer.ts#L1456-L1462) the comment _style_ hardcoded.

So, i'd plead you to refactor that part a bit so that the comment prefix (`//` in general vs `#` for coffeescript dialects) would be infered from the actual language being linted and not just assumed unconditionally to be `//` which in coffeescript's  case is definitively not true. 

thanks in advance and keep the awesome work :-) 

All the Best

António

helixbass:
@dbaeumer per #825 this PR makes the syntax for disabling comments language-specific so that when using the extension with Coffeescript source files it can generate the correct Coffeescript syntax for `eslint-disable*` comments

Fixes #825

Let me know if there's somewhere appropriate to add tests, I tested manually and saw the expected behavior for both Coffeescript and Javascript source files

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -1539,16 +1539,19 @@ messageQueue.registerRequest(CodeActionRequest.type, (params) => {
 		return TextEdit.replace(Range.create(textDocument!.positionAt(editInfo.edit.range[0]), textDocument!.positionAt(editInfo.edit.range[1])), editInfo.edit.text || '');
 	}
 
+	const languageSpecificLineCommentOpener = textDocument.languageId === 'coffeescript' ? '#' : '//';
+	const languageSpecificBlockCommentOpener = textDocument.languageId === 'coffeescript' ? '###' : '/*';
+	const languageSpecificBlockCommentCloser = textDocument.languageId === 'coffeescript' ? '###' : '*/';
 	function createDisableLineTextEdit(editInfo: Problem, indentationText: string): TextEdit {
-		return TextEdit.insert(Position.create(editInfo.line - 1, 0), `${indentationText}// eslint-disable-next-line ${editInfo.ruleId}${EOL}`);
+		return TextEdit.insert(Position.create(editInfo.line - 1, 0), `${indentationText}${languageSpecificLineCommentOpener} eslint-disable-next-line ${editInfo.ruleId}${EOL}`);
 	}
 
 	function createDisableSameLineTextEdit(editInfo: Problem): TextEdit {
-		return TextEdit.insert(Position.create(editInfo.line - 1, Number.MAX_VALUE), ` // eslint-disable-line ${editInfo.ruleId}`);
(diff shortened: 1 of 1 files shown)


AntonioMeireles:
@dbaeumer 

Hi, (curious), other than the rebase, what's holding this back ? (postponed to 2.2.x above)

Thanks.

dbaeumer:
Nothing. I needed to do a 2.1.0 release to fix some other things. This simply got pushed because of version numbering rules.

JamesHenry:
Hi @dbaeumer, please could I request that we rekindle this?

As soon as this lands in the Angular compiler: https://github.com/angular/angular/pull/41251

I can make it so that @angular-eslint users can use `eslint-disable` comments in their HTML. Naturally these need to be HTML comments `<!-- eslint-disable -->` and so the current behaviour in VSCode will yield broken results.

Please let me know if you would prefer I open a fresh PR, but this one seemed like it was already 90% of the way there for what we need for HTML.

dbaeumer:
Looking at this again the biggest problem is that the ESLint server shouldn't start to have this knowledge hard coded since we need to adjust this for every language that needs to be supported.

VS Code has this knowledge already in their language configuration but I am not sure if we are able to read it on the extension side. Will see if this is possible.

JamesHenry:
Thanks a lot, Dirk! And as always thank you for maintaining this extension 🙏

dbaeumer:
Actually there is an VS Code issue for this: https://github.com/microsoft/vscode/issues/2871

May be we should make this some sort of information in extension and send it as a config so that we later on can convert to some VS Code API.

JamesHenry:
FYI the feature is now available in angular-eslint, so using the VSCode fix on HTML files is now broken in the sense it puts in // comments instead of HTML ones

dbaeumer:
Closing the PR due to merge conflicts.

PR #925, opened 2020-03-17
Add markdown to eslint.probe default docs

btmills:
2e3cbb8 added markdown to the defaults for `eslint.probe`. This commit brings the documented default back in sync with the actual value.


dbaeumer:
Needs to wait until we publish the next version.

PR #946, opened 2020-04-15
Fixes #945: Handle null setting in isExplicitlyDisabled

### LINKED ISSUE #945: Error: "Cannot read property 'source.fixAll.eslint' of null"
This issue has come up with the latest version – `2.1.3`:

image

I reckon it's caused by the new `CodeActionsOnSave.isExplicitlyDisabled()` function here: https://github.com/microsoft/vscode-eslint/commit/5d3bc851de5a841ac90432e3bc95797c30eaf1da#diff-bc3c0f2fe3aca9ddbe96347ffb00641eR466-R471, which doesn't account for the case where `setting` is `null`.

Through no deliberate action of my own, my _User_-level settings were updated to include:
```
{
   // ...
   "editor.codeActionsOnSave": null
}
```

Again I'm unsure as to why this happened – it wasn't a manual edit. I am indeed told that `null` is an invalid value when viewing my settings file:

image

May be common enough to warrant a fix.

---
Issue is resolved by reverting to `2.1.2`.

andrao:
## Proposed changes
- In `CodeActionsOnSave. isExplicitlyDisabled()`, immediately return `false` when `setting` is falsy, as opposed to when `setting` is explicitly `undefined`.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -464,7 +464,7 @@ type CodeActionsOnSave = CodeActionsOnSaveMap | string[];
 
 namespace CodeActionsOnSave {
 	export function isExplicitlyDisabled(setting: CodeActionsOnSave | undefined): boolean {
-		if (setting === undefined || Array.isArray(setting)) {
+		if (!setting || Array.isArray(setting)) {
 			return false;
 		}
 		return setting['source.fixAll.eslint'] === false;


dbaeumer:
@andrao thanks for the PR. However this only handles the exception pass. I made some more changes to cover other passes as well. 

Could you still let me know how you got into the state of having `null` as the value?

Issue #983, opened 2020-05-29
Some eslint plugins crash without NODE_ENV


mattlubner:
Currently it's impossible to use vscode-eslint with any plugin that needs `NODE_ENV` because there's no way to define it.

The ideal solution would be to expose a `eslint.nodeEnv` setting, or a more generalized `eslint.processEnv` setting, so that users can define whatever environment variables are needed by their eslint plugins.

Example:

> [eslint-import-resolver-babel-module] Error: Using `babel-preset-razzle` requires that you specify `NODE_ENV` or `BABEL_ENV` environment variables. Valid values are "development", "test", and "production". Instead, received: undefined.

This is essentially the same issue as #741, but for some reason, that issue was closed when a platform-dependent workaround was found (which didn't address the underlying root cause). Clearly, defining `NODE_ENV` for the entire VS Code process is overkill and not really a viable solution, as doing so is likely to cause other problems.

dbaeumer:
Anyone interested working on a PR for this?

mattlubner:
I'm willing to take a stab at it, although I haven't worked with code in a VSCode extension before, or written much TypeScript either (just FYI).

Is this the right place to add environment variables? https://github.com/microsoft/vscode-eslint/blob/master/client/src/extension.ts#L886

Would these be the right places to add a new configuration setting?
- https://github.com/microsoft/vscode-eslint/blob/master/client/src/extension.ts#L181
- https://github.com/microsoft/vscode-eslint/blob/master/client/src/extension.ts#L1094

mattlubner:
PR is submitted, and changes seem to work correctly while running the debugging task! :)

mattlubner:
Still waiting on a final PR review + merge. Some attention on this issue would be really appreciated! 🙏😅🙏

dbaeumer:
Sorry. Slipped through the cracks. Will have a look.

dbaeumer:
I merged the PR. Will look into a new version when we ship a new version of VS Code beginning of August.

adriencaccia:
Thanks a lot @mattlubner, I'm using `eslint-import-resolver-babel-module` also with a custom `babel` config that requires a custom `NODE_ENV`, so this feature will gladly improve my workflow :smile:.

@dbaeumer, do you have an ETA for a new version of the extension containing this feature ?

Thanks again for all your work !

dbaeumer:
Actually this got delayed due to the security work. I will look into it in September.

PR #988, opened 2020-06-08
Add a NODE_ENV configuration setting for eslint

mattlubner:
Addresses #983

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -877,11 +877,13 @@ function realActivate(context: ExtensionContext): void {
 	const debug = eslintConfig.get('debug');
 	const nodeEnv = eslintConfig.get('nodeEnv', null);
 
-	const env: { [key: string]: string | number | boolean } | undefined = {};
+	let env: { [key: string]: string | number | boolean } | undefined;
 	if (nodeEnv) {
+		env = env || {};
 		env.NODE_ENV = nodeEnv;
 	}
 	if (debug) {
+		env = env || {};
 		env.DEBUG = 'eslint:*,-eslint:code-path';
 	}
 	const serverOptions: ServerOptions = {
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -878,14 +878,15 @@ function realActivate(context: ExtensionContext): void {
 	const debug = eslintConfig.get('debug');
 	const nodeEnv = eslintConfig.get('nodeEnv', null);
 
-	let env: { [key: string]: string | number | boolean } | undefined;
-	if (nodeEnv) {
-		env = env || {};
-		env.NODE_ENV = nodeEnv;
-	}
-	if (debug) {
-		env = env || {};
-		env.DEBUG = 'eslint:*,-eslint:code-path';
+	let env: { [key: string]: string | undefined } | undefined;
+	if (debug || nodeEnv) {
+		env = { ...process.env };
+		if (debug) {
+			env.DEBUG = 'eslint:*,-eslint:code-path';
+		}
+		if (nodeEnv) {
+			env.NODE_ENV = nodeEnv;
+		}
 	}
 	const serverOptions: ServerOptions = {
 		run: { module: serverModule, transport: TransportKind.ipc, runtime, options: { cwd: process.cwd(), env } },
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -178,6 +178,7 @@ interface ConfigurationSettings {
 	onIgnoredFiles: ESLintSeverity;
 	options: any | undefined;
 	run: RunValues;
+	nodeEnv: string | null;
 	nodePath: string | null;
 	workspaceFolder: WorkspaceFolder | undefined;
 	workingDirectory: ModeItem | DirectoryItem | undefined;
@@ -875,12 +876,14 @@ function realActivate(context: ExtensionContext): void {
 	const eslintConfig = Workspace.getConfiguration('eslint');
 	const runtime = eslintConfig.get('runtime', undefined);
 	const debug = eslintConfig.get('debug');
+	const nodeEnv = eslintConfig.get('nodeEnv', null);
 
-	let env: { [key: string]: string | number | boolean } | undefined;
+	const env: { [key: string]: string | number | boolean } | undefined = {};
+	if (nodeEnv) {
+		env.NODE_ENV = nodeEnv;
+	}
 	if (debug) {
-		env = {
-			DEBUG: 'eslint:*,-eslint:code-path'
-		};
+		env.DEBUG = 'eslint:*,-eslint:code-path';
 	}
 	const serverOptions: ServerOptions = {
 		run: { module: serverModule, transport: TransportKind.ipc, runtime, options: { cwd: process.cwd(), env } },
@@ -1091,6 +1094,7 @@ function realActivate(context: ExtensionContext): void {
 							onIgnoredFiles: ESLintSeverity.from(config.get<string>('onIgnoredFiles', ESLintSeverity.off)),
 							options: config.get('options', {}),
 							run: config.get('run', 'onType'),
+							nodeEnv: config.get('nodeEnv', null),
 							nodePath: config.get('nodePath', null),
 							workingDirectory: undefined,
 							workspaceFolder: undefined,
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -178,7 +178,6 @@ interface ConfigurationSettings {
 	onIgnoredFiles: ESLintSeverity;
 	options: any | undefined;
 	run: RunValues;
-	nodeEnv: string | null;
 	nodePath: string | null;
 	workspaceFolder: WorkspaceFolder | undefined;
 	workingDirectory: ModeItem | DirectoryItem | undefined;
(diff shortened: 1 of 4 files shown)


mattlubner:
Hey @dbaeumer, is there anything I can do to help expedite getting this PR reviewed / merged? 🙂

client/src/extension.ts
```diff
@@ -178,6 +178,7 @@ interface ConfigurationSettings {
 	onIgnoredFiles: ESLintSeverity;
 	options: any | undefined;
 	run: RunValues;
+	nodeEnv: string | null;
```
dbaeumer, line 181: This is not necessary since the value never travels to the server

client/src/extension.ts
```diff
@@ -875,12 +876,14 @@ function realActivate(context: ExtensionContext): void {
 	const eslintConfig = Workspace.getConfiguration('eslint');
 	const runtime = eslintConfig.get('runtime', undefined);
 	const debug = eslintConfig.get('debug');
+	const nodeEnv = eslintConfig.get('nodeEnv', null);
 
-	let env: { [key: string]: string | number | boolean } | undefined;
```
dbaeumer, line 879: We should only create an env if necessary to avoid copying the whole environment when starting the child process.

server/src/eslintServer.ts
```diff
@@ -198,6 +198,7 @@ interface CommonSettings {
 	onIgnoredFiles: ESLintSeverity;
 	options: ESLintOptions | undefined;
 	run: RunValues;
+	nodeEnv: string | null;
```
dbaeumer, line 201: This can be removed since the value is never used on the server.

client/src/extension.ts
mattlubner: I've made the change, but if my understanding of `LanguageClient` is correct (I'm assuming it passes `serverOptions.run.options` to `ChildProcess`), the whole environment wasn't being copied for either the `nodeEnv` or `debug` cases. Should it be?

client/src/extension.ts
mattlubner: Removed.

server/src/eslintServer.ts
mattlubner: Removed.

client/src/extension.ts
mattlubner: Looks like this was needed.

client/src/extension.ts
dbaeumer: We can also change the client code to check for an empty literal. The problem here is node API which doesn't automatically merge. So you need to create a new literal with the whole current env values.

client/src/extension.ts
dbaeumer: Can you explain why?

mattlubner: Removing it caused the test suite to fail, and the plugin to crash locally, due to a type failure.

mattlubner: (You should be able to see the exact error message in the failed Azure pipeline job.)

client/src/extension.ts
mattlubner: > We can also change the client code to check for an empty literal.

Do you mean changing `LanguageClient` itself? If so, that doesn't seem warranted. We'd lose the ability to set an empty environment, and I think it would be confusing to use.

> The problem here is node API which doesn't automatically merge.

_Exactly._

> So you need to create a new literal with the whole current env values.

Do you want the environment to be copied only when setting `NODE_ENV`? Or, should it be copied when  _any_ environment variable is set? I'm asking because the latter would be a behavioral change to how the `debug` setting currently works.

_To clarify, the environment isn't currently copied when the `DEBUG` environment variable is set; see: https://github.com/microsoft/vscode-eslint/blob/master/client/src/extension.ts#L881_

client/src/extension.ts
mattlubner: Direct link: https://dev.azure.com/ms/vscode-eslint/_build/results?buildId=87254&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=5caf77c8-9b10-50ef-b5c7-ca89c63e1c86&l=21

dbaeumer: I see but `nodeEnv` doesn't need to be on ConfigurationSettings since it is not traveling to the server. It is like debug that doesn't travel either. https://github.com/microsoft/vscode-eslint/blob/master/client/src/extension.ts#L877

mattlubner: Oh I see, I was passing it to the server on line 1100. Removed and working now.

client/src/extension.ts
mattlubner: I've copied the environment whenever `debug` or `nodeEnv` are set. Is this what you meant / wanted?

mattlubner:
@dbaeumer Could you review my changes when you get a chance?

I'm really eager to get these changes merged & published, as the only way I can use VSCode at work right now is via the `[Extension Development Host]` VSCode window spawned by the "Launch Client" launch configuration of this plugin. It works, but it feels silly! 😅

client/src/extension.ts
dbaeumer: Actually no. I was thinking about something like this:

```
		if (debug) {
                        env = env || {};
			env.DEBUG = 'eslint:*,-eslint:code-path';
		}
		if (nodeEnv) {
                        env = env || {};
			env.NODE_ENV = nodeEnv;
		}

```

mattlubner: Updated

mattlubner:
@dbaeumer Could you take a look? I believe I've made all of the requested changes.

mattlubner:
@dbaeumer Are we good to merge this?

mattlubner:
@dbaeumer @Hirse I noticed the repo ownership changed since I opened this PR… is there anything I can do to get this merged and released? As far as I can tell, it's good to go.

mattlubner:
@dbaeumer Sorry to keep nagging, but it's been over a month since I made the changes you requested! Did I miss anything? I'd really like to get this PR merged! 🙏🙏

mattlubner:
I've merged the latest upstream changes into the PR branch, and verified that everything is working as expected.

mattlubner:
I'd really appreciate some feedback regarding this PR!

By the way, the change requested is on a line that's unaffected by the actual change, which might be why GitHub isn't showing that PR comment as "outdated".

dbaeumer:
I merged it in but changed env back to `let env: { [key: string]: string | number | boolean }`. Was there a deeper reason why you remove number and boolean and added undefined. We should not allow undefined as a value in env when setting.

mattlubner:
Thanks!! It's been awhile, but I think I made that change because I saw a type error when running the extension in debug mode? I've gotten to know TypeScript a lot better since I wrote that though, I agree that including `undefined` was a mistake. Although NodeJs doesn't allow process.env to contain values other than `string` - such values are just coerced to string though, so it's not that big of a deal.

Thanks again!

ethanyou725:
@mattlubner would you mind provide a default value for `NODE_ENV` configuration ?
I have met an issue with CRA@5.0.0 and ejected
debug info:
```
2022-02-02T09:24:47.689Z eslint:cli-engine Lint /Users/yzg/workspace/test/react-rxjs/src/index.js
2022-02-02T09:24:47.689Z eslint:linter Linting code for /Users/yzg/workspace/test/react-rxjs/src/index.js (pass 1)
2022-02-02T09:24:47.689Z eslint:linter Verify
2022-02-02T09:24:47.689Z eslint:linter With ConfigArray: /Users/yzg/workspace/test/react-rxjs/src/index.js
2022-02-02T09:24:47.693Z eslint:linter Parsing error: [BABEL] /Users/yzg/workspace/test/react-rxjs/src/index.js: Using `babel-preset-react-app` requires that you specify `NODE_ENV` or `BABEL_ENV` environment variables. Valid values are "development", "test", and "production". Instead, received: undefined. (While processing: "/Users/yzg/workspace/test/react-rxjs/node_modules/babel-preset-react-app/index.js")
Error: [BABEL] /Users/yzg/workspace/test/react-rxjs/src/index.js: Using `babel-preset-react-app` requires that you specify `NODE_ENV` or `BABEL_ENV` environment variables. Valid values are "development", "test", and "production". Instead, received: undefined. (While processing: "/Users/yzg/workspace/test/react-rxjs/node_modules/babel-preset-react-app/index.js")
    at module.exports (/Users/yzg/workspace/test/react-rxjs/node_modules/babel-preset-react-app/create.js:58:11)
    at module.exports (/Users/yzg/workspace/test/react-rxjs/node_modules/babel-preset-react-app/index.js:19:10)
    at sync (/Users/yzg/workspace/test/react-rxjs/node_modules/@babel/core/lib/gensync-utils/async.js:38:25)
    at sync (/Users/yzg/workspace/test/react-rxjs/node_modules/gensync/index.js:182:19)
    at /Users/yzg/workspace/test/react-rxjs/node_modules/gensync/index.js:210:24
    at Generator.next (<anonymous>)
    at /Users/yzg/workspace/test/react-rxjs/node_modules/@babel/core/lib/config/full.js:217:21
    at Generator.next (<anonymous>)
    at Function.<anonymous> (/Users/yzg/workspace/test/react-rxjs/node_modules/@babel/core/lib/gensync-utils/async.js:25:3)
    at Generator.next (<anonymous>)
```
https://github.com/facebook/create-react-app/blob/main/packages/babel-preset-react-app/create.js#L57
if there is no `NODE_ENV` setted, it will throw an error, I think set to `NODE_ENV=development` as default  might fix this.

Issue #1003, opened 2020-07-02
Feature Request: Can we make the server npm-installable?


justinabrahms:
I'd love the ability to start the server with something like this:

```
npm install vscode-eslint
vscode-eslint-server
```

This will allow me to use this language server in other proceses (beyond eslint) with easy installation mechanisms.

dbaeumer:
Agree that this would be cool.

danielpza:
@dbaeumer any news on this issue? I'll be glad to work on this feature if needed.

dbaeumer:
No news. Feature is marked as `help wanted`

danielpza:
@dbaeumer I made a PR #1053

dbaeumer:
Commented on the PR.

dbaeumer:
The feature has not gain any traction in the community. I therefore close the issue. Still a big Thank You to you for taking the time to create this issue! To learn more about how we handle feature requests, please see our [documentation](https://aka.ms/vscode-issue-lifecycle).

Happy Coding!

entropitor:
This comment is probably very relevant here as well: https://github.com/microsoft/vscode-eslint/pull/1053#issuecomment-936237529

PR #1053, opened 2020-08-26
Prepare server to deploy as a separate package

danielpza:
Installing `eslint-server` would start the eslint standalone server. It will need to be published to the npm repository:

```sh
cd server
npm install
npm publish # or --dry-run to check everything is ok
```

You can try it out by running `npm link` inside the server/ directory, it will link the command to the global path.

You can change the name of the package/command by modifying the `name` property of `server/package.json`, I would suggest `eslint-language-server` or something like that

Address #1003


dbaeumer:
@danielpza beside publishing it to npm I think we would need a documentation of the private non LSP message the ESLint server uses / provides. Otherwise it will be hard to talk to the server. In addition we need to decide how we deal with versioning if the server uses new private messages.

rchl:
Here is a rundown of custom notifications/requests:

Notifications from the server:
 - `eslint/status`
 - `eslint/exitCalled`

Requests from the server:
 - `eslint/openDoc`
 - `eslint/noConfig`
 - `eslint/noLibrary`
 - `eslint/probeFailed`
 - `eslint/confirmLocalESLint`

<hr>

There is some custom handling needed when the client is responding to `workspace/configuration`:
 - The `workspaceFolder` needs to be set for files that are within the workspace
 - If the client/user has defined `workingDirectories` then this setting needs to resolved into a value that is set as `workingDirectory` property.

<hr>

The `probe` configuration option also needs to be resolved and the resolved value set as `validate` configuration.

This is optional but it provides a better experience if implemented.

<hr>

So basically all of this would need to be documented as it's a functionality supported by the server but currently handled within the extension so it will be missing when using server with another editor.

danielpza:
I included some docs in the PR, however I don't know much about the workings of the server to document them all. I think it is a good start, the other clients have been using the server without any docs so I think it will be fine and further docs can be included later.

yyoncho:
> @danielpza beside publishing it to npm I think we would need a documentation of the private non LSP message the ESLint server uses / provides. Otherwise it will be hard to talk to the server. In addition we need to decide how we deal with versioning if the server uses new private messages.

@dbaeumer thank you for that. I just want to mention that the majority of the servers(with few exceptions) do not document these methods and we(at emacs side) are used to reverse-engineering the extension methods.

dbaeumer:
For documenting the configuration request. What needs to go into the document is not that the request is send. What we need to document is the structure of the response. (e.g. this type https://github.com/microsoft/vscode-eslint/blob/master/server/src/eslintServer.ts#L217)

danielpza:
@dbaeumer I think we should leave the documentation for another PR or for the wiki. We are missing out on the standalone server

dbaeumer:
I disagree. The reason is simple: not having the documentation will result in additional issue that will come via this repository. So I would really like to see the documentation  together with the npm package.

danielpza:
What additional issues do you expect to come to the repository? If they are documentation related it will be a good way to expand the documentation on demand.

dbaeumer:
People asking for the documentation that someone then has to write.

Sebastianokunkelo:
danielpza:prepare-server-deploy

williamboman:
FWIW, this repo is packaged via https://github.com/hrsh7th/vscode-langservers-extracted as of recently. Obviously not an official package, but thought I'd mention it

PR #1084, opened 2020-10-08
support remote development by indicating this is a workspace-only extension

bradzacher:
See docs: https://code.visualstudio.com/api/advanced-topics/remote-extensions#incorrect-execution-location


dbaeumer:
Makes sense. Thanks.

PR #993, opened 2020-06-14
Implement "eslint.rules.customizations" - Allow users to override rule severity from VS Code settings

jamesorlakin:
This hopefully closes #468 based on the discussions in #561.

This PR implements a new extension option, allowing diagnostics reported by the ESLint extension to be overridden by user settings within VS Code. The main usecase is to allow linting errors to be visually different to actual typing/compilation problems based on personal preference, especially if adjusting the ESLint configuration itself is difficult (e.g. it's prepackaged, StandardJS, etc) or there's a fear of committing these customizations accidentally.

For example, the playground environment in this PR:
image
image
image

The `eqeqeq` rule is an error in ESLint parlance, but the `*` override forces it to be a nicer yellow warning. I personally prefer this representation of all linting warnings.

---

All comments are welcome.

PR #1164, opened 2021-01-23
Implement eslint.rules.customizations - with overrides

### LINKED ISSUE #468: Feature Request: "Global Rule Severity" Setting

### LINKED ISSUE #1199: Reduce each severity by one level, i.e. display ESLint error as VS Code warning and display ESLint warning as VS Code info
**Issue**
As pointed out in issue #468, most ESLint shared configs use "error" as their default rule severity, but should often be displayed as warnings in VS Code since they are often less important and thus should not be mixed with TypeScript errors. In addition, an even less important (e.g. stylistic) rule in ESLint is normally (or can be) configured with severity "warn", which should be displayed as "Information" in VS  Code, to be  distinguishable from more important ESLint errors (displayed as VS Code warnings).

**Proposed Solution**
Add an configuration flag to reduce each severity by one level, i.e. display an ESLint error as VS Code warning and display an ESLint warning as VS Code information.

In Pull Request #1164 (authored by @JoshuaKGoldberg), a solution to this issue is proposed, but it seems to need more configuration. Another disadvantage of this "duplicated" configuration is, that it does not allow to re-use the difference in the configuration of the severity levels for build scripts. I.e. a warning problem in VS Code should probably fail an integration build, because there are ESLint errors, but information problems in VS Code should never fail the build, because they are configured as ESLint warnings.

JoshuaKGoldberg:
Continues @jamesorlakin's great work from https://github.com/microsoft/vscode-eslint/pull/993, with the changes requested by @dbaeumer to take settings as an array and allow ignoring rules.

Fixes #468 based on the discussions in #561.

Fixes #1199.

In the demo, note the underline under `bar` and `console.log` changing to red (error):

https://user-images.githubusercontent.com/3335181/110977314-0924af00-8330-11eb-8142-79b6fc8e85c9.mov

Co-authored-by: James Lakin <contact@jameslakin.co.uk>

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ This extension contributes the following variables to the [settings](https://cod
 - `eslint.rules.customizations`: force rules to report a different severity within VS Code compared to the project's true ESLint configuration. contain two properties:
   - `"rule`": Select on rules with names that match, factoring in asterisks as wildcards: `{ "rule": "no-*", "severity": "warn" }`
     - Prefix the query with a `"!"` to target all rules that _don't_ match the query: `{ "rule": "!no-*", "severity": "info" }`
-  - `"severity"`: Sets a new severity for matched rule(s), `"downgrade"`s them to a lower severity, `"upgrade"`s them to a higher severity, or `"reset"`s them to their original severity
+  - `"severity"`: Sets a new severity for matched rule(s), `"downgrade"`s them to a lower severity, `"upgrade"`s them to a higher severity, or `"default"`s them to their original severity
 
   In this example, all rules are overridden to warnings:
 
@@ -159,13 +159,13 @@ This extension contributes the following variables to the [settings](https://cod
   ]
   ```
 
-  In this example, `no-` rules are informative, other rules are downgraded, and `"radix"` is reset:
+  In this example, `no-` rules are informative, other rules are downgraded, and `"radix"` is reset to default:
 
   ```json
   "eslint.rules.customizations": [
     { "rule": "no-*", "severity": "info" },
     { "rule": "!no-*", "severity": "downgrade" },
-    { "rule": "radix", "severity": "reset" }
+    { "rule": "radix", "severity": "default" }
   ]
   ```
 
--- a/README.md
+++ b/README.md
@@ -146,9 +146,9 @@ This extension contributes the following variables to the [settings](https://cod
   - `all`: fixes all possible problems by revalidating the file's content. This executes the same code path as running eslint with the `--fix` option in the terminal and therefore can take some time. This is the default value.
   - `problems`: fixes only the currently known fixable problems as long as their textual edits are non overlapping. This mode is a lot faster but very likely only fixes parts of the problems.
 
-- `eslint.rules.customizations`: force rules to report a different severity within VS Code compared to the project's true ESLint configuration. contain two properties:
+- `eslint.rules.customizations` (@since 2.1.20): force rules to report a different severity within VS Code compared to the project's true ESLint configuration. Contains two properties:
   - `"rule`": Select on rules with names that match, factoring in asterisks as wildcards: `{ "rule": "no-*", "severity": "warn" }`
-    - Prefix the query with a `"!"` to target all rules that _don't_ match the query: `{ "rule": "!no-*", "severity": "info" }`
+    - Prefix the name with a `"!"` to target all rules that _don't_ match the name: `{ "rule": "!no-*", "severity": "info" }`
   - `"severity"`: Sets a new severity for matched rule(s), `"downgrade"`s them to a lower severity, `"upgrade"`s them to a higher severity, or `"default"`s them to their original severity
 
   In this example, all rules are overridden to warnings:
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ This extension contributes the following variables to the [settings](https://cod
 
 - `eslint.rules.customizations`: force rules to report a different severity within VS Code compared to the project's true ESLint configuration. This is an array that allows two kinds of glob patterns:
   - `"override`": Overrides for rules with names that match, factoring in asterisks: `{ "override": "no-*", "severity": "warn" }`
-  - `"ignore"`: Excludes rules matching a glob from previous overrides: `{ "ignore": "*jsx*" }`
+  - `"reset"`: Excludes rules matching a glob from previous overrides: `{ "reset": "*jsx*" }`
 
   In this example, all rules are overridden to warnings:
 
@@ -164,8 +164,8 @@ This extension contributes the following variables to the [settings](https://cod
   "eslint.rules.customizations": [
     { "override": "*", "severity": "warn" },
     { "override": "no-*", "severity": "info" },
-    { "ignore": "*semi*" },
-    { "ignore": "radix" }
+    { "reset": "*semi*" },
+    { "reset": "radix" }
   ]
   ```
 
--- a/README.md
+++ b/README.md
@@ -146,9 +146,10 @@ This extension contributes the following variables to the [settings](https://cod
   - `all`: fixes all possible problems by revalidating the file's content. This executes the same code path as running eslint with the `--fix` option in the terminal and therefore can take some time. This is the default value.
   - `problems`: fixes only the currently known fixable problems as long as their textual edits are non overlapping. This mode is a lot faster but very likely only fixes parts of the problems.
 
-- `eslint.rules.customizations`: force rules to report a different severity within VS Code compared to the project's true ESLint configuration. This is an array that allows two kinds of glob patterns:
-  - `"override`": Overrides for rules with names that match, factoring in asterisks: `{ "override": "no-*", "severity": "warn" }`
-  - `"reset"`: Excludes rules matching a glob from previous overrides: `{ "reset": "*jsx*" }`
+- `eslint.rules.customizations`: force rules to report a different severity within VS Code compared to the project's true ESLint configuration. contain two properties:
+  - `"rule`": Select on rules with names that match, factoring in asterisks as wildcards: `{ "override": "no-*", "severity": "warn" }`
+    - Prefix the query with a `"!"` to target all rules that _don't_ match the query: `{ "override": "!no-*", "severity": "info" }`
+  - `"override"`: Sets a new severity for matched rule(s), `"downgrade"`s them to a lower severity, `"upgrade"`s them to a higher severity, or `"reset"`s them to their original severity
 
   In this example, all rules are overridden to warnings:
 
@@ -158,14 +159,13 @@ This extension contributes the following variables to the [settings](https://cod
   ]
   ```
 
-  In this example, with the exception of `semi` rules and `radix`, all all `no-` rules are warnings and other rules are info:
+  In this example, `no-` rules are informative, other rules are downgraded, and `"radix"` is reset:
 
   ```json
   "eslint.rules.customizations": [
-    { "override": "*", "severity": "warn" },
     { "override": "no-*", "severity": "info" },
-    { "reset": "*semi*" },
-    { "reset": "radix" }
+    { "override": "!no-*", "severity": "downgrade" },
+    { "reset": "radix", "severity": "reset" }
   ]
   ```
 
--- a/README.md
+++ b/README.md
@@ -146,6 +146,29 @@ This extension contributes the following variables to the [settings](https://cod
   - `all`: fixes all possible problems by revalidating the file's content. This executes the same code path as running eslint with the `--fix` option in the terminal and therefore can take some time. This is the default value.
   - `problems`: fixes only the currently known fixable problems as long as their textual edits are non overlapping. This mode is a lot faster but very likely only fixes parts of the problems.
 
+- `eslint.rules.customizations`: force rules to report a different severity within VS Code compared to the project's true ESLint configuration. This is an array that allows two kinds of glob patterns:
+  - `"override`": Overrides for rules with names that match, factoring in asterisks: `{ "override": "no-*", "severity": "warn" }`
+  - `"ignore"`: Excludes rules matching a glob from previous overrides: `{ "ignore": "*jsx*" }`
+
+  In this example, all rules are overridden to warnings:
+
+  ```json
+  "eslint.rules.customizations": [
+    { "override": "*", "severity": "warn" }
+  ]
+  ```
+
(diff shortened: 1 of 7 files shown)


server/src/eslintServer.ts
```diff
@@ -368,7 +386,27 @@ function loadNodeModule<T>(moduleName: string): T | undefined {
 	return undefined;
 }
 
-function makeDiagnostic(problem: ESLintProblem): Diagnostic {
+function asteriskMatches(matcher: string, ruleId: string) {
+	return new RegExp(`^${matcher.replace(/\*/g, '.*')}$`, 'g').test(ruleId);
```
JoshuaKGoldberg, line 390: This could become a performance concern with many overrides and many errors to report. In theory we could keep a cache of matchers to rule IDs, but I'm not familiar with VS Code extension performance concerns. Would you like me to do that? If so, would you like me to make it only keep, say, the most recent 100 overrides, to keep memory low?

client/src/extension.ts
```diff
@@ -181,6 +181,23 @@ enum ConfirmationSelection {
...
+type RuleCustomization = RuleIgnore | RuleOverride;
+
+interface RuleIgnore {
+	ignore: string;
+}
+
+interface RuleOverride {
+	override: string;
+	severity: RuleSeverity;
+}
```
JoshuaKGoldberg, line 199: It's a little unfortunate typing these out twice. Would you like me to set up a `shared` folder & tsconfig project to keep these as types-only constructs?

server/src/eslintServer.ts
dbaeumer: We should definitely try to address this. Can we no cache this from RuleId to new severity. Usually this doesn't change a lot for files during validation. A LRUCache implementation can be borrowed from here: https://github.com/microsoft/vscode-languageserver-node/blob/master/jsonrpc/src/common/linkedMap.ts#L6

dbaeumer (changes requested):

server/src/eslintServer.ts
JoshuaKGoldberg: Thanks for the pointer! I tried it out and it worked well -- though I'm wondering if my initial suggestion of an LRU cache might be more overhead than necessary? I skimmed through some larger ESLinted codebases and didn't see any that use more than a couple hundred rules at most. 
* bbc3eb0 shows using a basic `Map<string, RuleSeverity | undefined>`
* 0cc0a8a (current) switches that to a `LinkedMap`

I suspect the plain `Map` is sufficient but defer to you. 😄

dbaeumer: I see your point but I usually stay away from unbounded caches since they open an attack vector. Given that ESLint is executed in the cloud as well when running VS Code remote / code space I would opt to use the LRUCache from the linkedMap.ts file. This will allow us to limit the cache to lets say 1024 entries.

JoshuaKGoldberg: Very interesting, TIL. Thank you for the explanation!

dbaeumer: @JoshuaKGoldberg please ping when you think the code is ready for another review.

JoshuaKGoldberg: Ah apologies, I didn't want to spam. @dbaeumer the code is ready for another review.

client/src/extension.ts
dbaeumer: Having them in both client and server is fine with me. We have that for other settings as well. We can always move them to a common folder and link that folder so that the resuling js files can simple be created twice for client & server.

See my additional comments why these can't be type only

README.md
```diff
@@ -146,6 +146,29 @@ This extension contributes the following variables to the [settings](https://cod
   - `all`: fixes all possible problems by revalidating the file's content. This executes the same code path as running eslint with the `--fix` option in the terminal and therefore can take some time. This is the default value.
   - `problems`: fixes only the currently known fixable problems as long as their textual edits are non overlapping. This mode is a lot faster but very likely only fixes parts of the problems.
 
+- `eslint.rules.customizations`: force rules to report a different severity within VS Code compared to the project's true ESLint configuration. This is an array that allows two kinds of glob patterns:
+  - `"override`": Overrides for rules with names that match, factoring in asterisks: `{ "override": "no-*", "severity": "warn" }`
+  - `"reset"`: Excludes rules matching a glob from previous overrides: `{ "reset": "*jsx*" }`
```
dbaeumer, line 151: Can you please provide an example on when users should use that.

client/src/extension.ts
```diff
@@ -1441,6 +1459,7 @@ function realActivate(context: ExtensionContext): void {
 							quiet: config.get('quiet', false),
 							onIgnoredFiles: ESLintSeverity.from(config.get<string>('onIgnoredFiles', ESLintSeverity.off)),
 							options: config.get('options', {}),
+							rulesCustomizations: config.get('rules.customizations', []),
```
dbaeumer, line 1462: We need to sanitize these. Users can type whatever they want into the settings.json so we can't blindly take the values. See ESLintSeverity.

And that opens the next questions: IMO RuleSeverity and ESLintSeverity should be folded into one.

playgrounds/6.0/.vscode/settings.json
```diff
@@ -39,6 +39,13 @@
 		"location": "separateLine"
 	},
 	"eslint.quiet": false,
+	"eslint.rules.customizations": [
+		{ "override": "*", "severity": "info" },
+		{ "override": "no-*", "severity": "warn" },
+		{ "reset": "*console*" },
```
dbaeumer, line 45: I haven't fully thought they through but would it be better to have a negative pattern like in .eslintignore or gitignore. Somethings like `"override": "!*console*"`. And then we could name `override` -> `rule`

server/src/eslintServer.ts
```diff
@@ -1290,6 +1348,12 @@ function validate(document: TextDocument, settings: TextDocumentSettings & { lib
 		}
 	}
 
+	const newRuleCustomizationsKey = JSON.stringify(settings.rulesCustomizations);
```
dbaeumer, line 1351: I would actually hash this to make it clear that we use the content as a key.

dbaeumer (changes requested):

dbaeumer:
Add my comment. Again thanks for working on this.

dbaeumer:
@JoshuaKGoldberg anything I need to do to move this forward?

JoshuaKGoldberg:
Ah! It fell off my radar. Looking again now -- thanks for the ping!

server/src/eslintServer.ts
JoshuaKGoldberg: Err, is there a particular hash function you're looking for here? I'm under the impression `JSON.stringify` works as a rudimentary, though space-inefficient, hash.

playgrounds/6.0/.vscode/settings.json
JoshuaKGoldberg: 👍 . Putting this forward in https://github.com/microsoft/vscode-eslint/issues/1199#issuecomment-797526235 before implementing here, in case there's more churn.

stenzengel:
Can I help? E.g. is it easily possible to test this PR before it is merged or a new version is released? (Unfortunately I have no experience with VS Code extension development, yet.)

dbaeumer:
@stenzengel  To test things you can do the following:

- clone @JoshuaKGoldberg fork
- npm install
- open in VS Code
- go to the debugger tab
- run Launch Client

That launches a new VS Code instance which uses the ESLint extension from source.

stenzengel:
@JoshuaKGoldberg 
I have only done a quick test, but my use case works like a charm so far. I will continue testing.

One short question: I'm not sure, if you prefer the property name "override" to my original proposal "severity" or if it is an oversight. I am not a native speaker, so I am not sure which one sounds better.

Current implementation:
```jsonc
{ "rule": "*",         "override": "downgrade" },
{ "rule": "*console*", "override": "reset" },
```
Original proposal
```jsonc
{ "rule": "*",         "severity": "downgrade" },
{ "rule": "*console*", "severity": "reset" },
```

dbaeumer:
I prefer `severity` over `override`

JoshuaKGoldberg:
This is how it is now:

```json
    { "override": "no-*", "severity": "info" },
    { "override": "!no-*", "severity": "downgrade" },
    { "reset": "radix", "severity": "reset" }
  ]
```

Is that acceptable to both of you?

stenzengel:
I don't understand yet in which case exactly you have to use the property **name** "reset" instead of "override". Only in cases where the **value** for "severity" is also "reset"?. I also don't see any commit yet where I could hava a closer look.

I definitely don't want you to make a change to your code that you're not completely happy with because of my configuration proposition.  I.e. for me all configuration alternatives are ok as long as it is possible to downgrade all rules.

JoshuaKGoldberg:
Oh you know what, I had just forgotten to update the README.md example 😅 . Sorry! Yes, I'll switch to this -- your proposal @stenzengel:

```jsonc
    { "rule": "no-*", "severity": "info" },
    { "rule": "!no-*", "severity": "downgrade" },
    { "rule": "radix", "severity": "reset" }
  ]
```

dbaeumer:
One minor suggestion: I would like to rename `reset` since to my undestanding it will use the value defined in the eslintrc file. Suggestions are: `default` since it is the value in the `eslintrc` file or `eslintrc` to make this even more clear. I actually woul prefer default.

dbaeumer:
@JoshuaKGoldberg is this ready for another review?

JoshuaKGoldberg:
@dbaeumer yes please!

README.md
```diff
@@ -146,6 +146,29 @@ This extension contributes the following variables to the [settings](https://cod
   - `all`: fixes all possible problems by revalidating the file's content. This executes the same code path as running eslint with the `--fix` option in the terminal and therefore can take some time. This is the default value.
   - `problems`: fixes only the currently known fixable problems as long as their textual edits are non overlapping. This mode is a lot faster but very likely only fixes parts of the problems.
 
+- `eslint.rules.customizations`: force rules to report a different severity within VS Code compared to the project's true ESLint configuration. contain two properties:
```
dbaeumer, line 149: contain -> Contain

README.md
```diff
@@ -146,6 +146,29 @@ This extension contributes the following variables to the [settings](https://cod
   - `all`: fixes all possible problems by revalidating the file's content. This executes the same code path as running eslint with the `--fix` option in the terminal and therefore can take some time. This is the default value.
   - `problems`: fixes only the currently known fixable problems as long as their textual edits are non overlapping. This mode is a lot faster but very likely only fixes parts of the problems.
 
+- `eslint.rules.customizations`: force rules to report a different severity within VS Code compared to the project's true ESLint configuration. contain two properties:
+  - `"rule`": Select on rules with names that match, factoring in asterisks as wildcards: `{ "rule": "no-*", "severity": "warn" }`
+    - Prefix the query with a `"!"` to target all rules that _don't_ match the query: `{ "rule": "!no-*", "severity": "info" }`
```
dbaeumer, line 151: I would rename query with name

README.md
dbaeumer, line 149: We should add a @since tag like with other new settings.

client/src/extension.ts
```diff
@@ -399,6 +417,23 @@ function computeValidate(textDocument: TextDocument): Validate {
 	return Validate.off;
 }
 
+function parseRulesCustomizations(rawConfig: unknown): RuleCustomization[] {
+	if (!rawConfig || !Array.isArray(rawConfig)) {
+		return [];
+	}
+
+	return rawConfig.map(rawValue => {
+		if ('severity' in rawValue && typeof rawValue.severity === 'string' && 'rule' in rawValue && typeof rawValue.rule === 'string') {
```
dbaeumer, line 426: We do we need the `in` check. If the value is not a property in rawValue then the typeof check will fail.

dbaeumer (changes requested):
Very nice work. Only minor things. Thanks again for doing this.

dbaeumer:
Merged !!

dbaeumer:
I drafted a pre-release here: https://github.com/microsoft/vscode-eslint/releases/tag/insider%2F2.1.20 in case you want to give it a try.

JoshuaKGoldberg:
Whoohoo! 🙌 

Thanks everyone for helping push this through! It feels like we made a real team effort and I'm stoked to start adding this config to repositories. 🚀

stenzengel:
Thanks a lot, @JoshuaKGoldberg and @dbaeumer . The pre-release worked for me without any problems.

lolmaus:
Hi!

I have installed v2.1.20, added this to my `settings.json`:

```json
        {
            "rule": "*", "override": "info"
        }
    ],
```

I expect the linting problems to be underlined yellow, but they are still red. What am I doing wrong?

dbaeumer:
@lolmaus the correct syntax is:

```
		{ "rule": "*", "severity": "info" }
	]
```

lolmaus:
Ah, thanks! My syntax is correct, but I used the keyword `override` from this PR's screenshot. The correct keyword is `severity`.

Thank you for the *absolutely awesome* feature! 🙇‍♂️🙏

lolmaus:
Here's a config that sets the color of fixable rules to blue, non-fixable to yellow:
https://gist.github.com/lolmaus/e1be5655288fca67e53a22b720adf40d

Includes `eslint`, `@typescript-eslint` and `prettier`.

lolmaus:
image

I love how different severities overlap and you can now see an important problem inside an unimportant problem! 😍

PR #1192, opened 2021-03-02
adopt backgroundColor option for status bar item

bpasero:
I cannot stand seeing this anymore:

image


dbaeumer:
@bpasero thanks. Was on my list but I didn't get to it in the last milestone.

PR #1226, opened 2021-03-28
feat(client): automatically detect package manager

### LINKED ISSUE #1141: Automatically detect package manager
Based on https://github.com/microsoft/vscode/issues/102050

VSCode now automatically detects the package manager by default and exposes the package manager (whether automatically discovered or manually specified in `npm.packageManager` setting) in the `npm.packageManager` command. As a result we can use the `npm.packageManager` command instead of our own `packageManager` setting.

There's a similar implementation in https://github.com/prettier/prettier-vscode/pull/1700, but this extension has a different architecture and I couldn't figure out how to access the package manager from the language server.

hantatsang:
Close #1141

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -181,9 +181,11 @@ enum ConfirmationSelection {
 	alwaysAllow = 4
 }
 
+type NpmPackageManager = 'npm' | 'pnpm' | 'yarn';
+
 interface ConfigurationSettings {
 	validate: Validate;
-	packageManager: 'npm' | 'yarn' | 'pnpm';
+	packageManager: NpmPackageManager | undefined;
 	codeAction: CodeActionSettings;
 	codeActionOnSave: CodeActionsOnSaveSettings;
 	format: boolean;
@@ -1550,9 +1552,11 @@ function realActivate(context: ExtensionContext): void {
 								}
 							}
 						});
+						const packageManager = await Commands.executeCommand<NpmPackageManager>('npm.packageManager', workspaceFolder?.uri);
 						const settings: ConfigurationSettings = {
 							validate: Validate.off,
-							packageManager: config.get('packageManager', 'npm'),
+
+							packageManager: !!packageManager ? packageManager : undefined,
 							codeActionOnSave: {
 								enable: false,
 								mode: CodeActionsOnSaveMode.all
@@ -1777,7 +1781,7 @@ function realActivate(context: ExtensionContext): void {
 
 			const uri: Uri = Uri.parse(params.source.uri);
 			const workspaceFolder = Workspace.getWorkspaceFolder(uri);
-			const packageManager = Workspace.getConfiguration('eslint', uri).get('packageManager', 'npm');
+			const packageManager = Workspace.getConfiguration('eslint', uri).get<NpmPackageManager>('packageManager', 'npm');
 			const localInstall = {
 				npm: 'npm install eslint',
 				pnpm: 'pnpm install eslint',
@@ -1788,7 +1792,6 @@ function realActivate(context: ExtensionContext): void {
 				pnpm: 'pnpm install -g eslint',
 				yarn: 'yarn global add eslint'
 			};
-			const isPackageManagerNpm = packageManager === 'npm';
 			interface ButtonItem extends MessageItem {
 				id: number;
 			}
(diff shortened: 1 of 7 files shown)


dbaeumer:
@hantatsang thanks for the PR. I would like to ask you to check one thing. The npm.packageManager setting is contributed by the build in NPM extension, which a user can disable. What will happen if a user has disabled the extension?

hantatsang:
@dbaeumer That's a good point. 
In that case `ConfigurationSettings.packageManager` is fallback to `npm`. Then when `eslint` is installed in:

- Local `npm` or `pnpm` or `yarn`: `eslint` loaded

    ```
    [Trace - 9:12:11 am] Received notification 'window/logMessage'.
    Params: {
        "type": 3,
        "message": "ESLint library loaded from: D:\\vscode-eslint\\1141-auto-detect-package-manager\\node_modules\\eslint\\lib\\api.js"
    }
    ```

- Global `npm`: `eslint` loaded

    ```
    [Trace - 9:07:12 am] Received notification 'window/logMessage'.
    Params: {
        "type": 3,
        "message": "ESLint library loaded from: C:\\ProgramData\\nvm\\v15.12.0\\node_modules\\eslint\\lib\\api.js"
    }
    ```

- Global `pnpm` or `yarn`: `eslint` can't be loaded. But same behaviour observed for `master` branch - no linting happens.

client/src/extension.ts
```diff
@@ -1550,9 +1552,10 @@ function realActivate(context: ExtensionContext): void {
 								}
 							}
 						});
+						const packageManager = await Commands.executeCommand<NpmPackageManager>('npm.packageManager', workspaceFolder?.uri);
```
hantatsang, line 1555: Need `try`/`catch` to handle `UnhandledPromiseRejectionWarning` error when NPM is disabled

dbaeumer:
OK. We can do the following. We can keep the setting but first see if it is set. If not use `npm.packageManager` otherwise `eslint.packageManager`. Does that make sense?

hantatsang:
Yep sounds good. I've made the update.

dbaeumer:
I am a little pluzzled by the descriptions of the options you put in. Should we simply say that if the setting is not provided the package manager defaults to `npm.packageManger`. Why do we need to point out the global case?

hantatsang:
Sorry, looks like I got myself confused. How about now?

client/src/extension.ts
```diff
@@ -1824,7 +1835,6 @@ function realActivate(context: ExtensionContext): void {
 				client.info([
 					`Failed to load the ESLint library for the document ${uri.fsPath}`,
 					`To use ESLint for single JavaScript file install eslint globally using '${globalInstall[packageManager]}'.`,
-					isPackageManagerNpm ? 'If you are using yarn or pnpm instead of npm set the setting `eslint.packageManager` to either `yarn` or `pnpm`' : null,
```
dbaeumer, line 1827: Why do we remove this. We should still point out that if you use a different package manager than npm that there are now to settings to control the package manager to use.

dbaeumer (changes requested):
See my comment about the removal of the line

```
isPackageManagerNpm ? 'If you are using yarn or pnpm instead of npm set the setting `eslint.packageManager` to either `yarn` or `pnpm`' : null,
```

client/src/extension.ts
```diff
@@ -1816,6 +1817,7 @@ function realActivate(context: ExtensionContext): void {
 					`To use ESLint please install eslint by running ${localInstall[packageManager]} in the workspace folder ${workspaceFolder.name}`,
 					`or globally using '${globalInstall[packageManager]}'. You need to reopen the workspace after installing eslint.`,
 					'',
+					isPackageManagerNpm ? 'If you are using yarn or pnpm instead of npm set the setting `eslint.packageManager` or `npm.packageManager` to either `yarn` or `pnpm`' : null,
```
dbaeumer, line 1820: One minor thing. Since we want users to use `npm.packageManager` I would name that first.

dbaeumer (changes requested):

hantatsang:
Is this good to go yet @dbaeumer ?

dbaeumer:
I was busy with non ESLint stuff in May. Will have a look in June.

Issue #1270, opened 2021-05-25
eslint doesn't display any error in ts file


ananiy:
[Archive.zip](https://github.com/microsoft/vscode-eslint/files/6537496/Archive.zip)

my node_modules directory Is generated by tnpm, which is my company developed. 

my question is, I can't see any error in ts file. 

what's wrong with this, plz help me🙏

dbaeumer:
@ananiy what happens if you try to validate files in the terminal?

ananiy:
> @ananiy what happens if you try to validate files in the terminal?

it works well. so I think some problem happens in the extension. 

you can download the zip file to try, then delete node_modules dir and `npm i`, it's easy to replay.

ananiy:
when using tnpm: 
![222](https://user-images.githubusercontent.com/33290601/119591388-128ac700-be09-11eb-9297-b4832ad1bc8e.jpg)
[222.txt](https://github.com/microsoft/vscode-eslint/files/6543360/222.txt)

when using npm: 
![111](https://user-images.githubusercontent.com/33290601/119591509-4fef5480-be09-11eb-93d5-733d5fe5543c.jpg)
[111.txt](https://github.com/microsoft/vscode-eslint/files/6543361/111.txt)

ananiy:
> @ananiy now I see. The extension has no official support for `tnpm`.

but tnpm is my company inside version, It can not install from npm registry.
I can only provide the compressed node_modules directory. 
what else would I support? 
thx :)

dbaeumer:
Which version of `tnpm` are you using and where can I find it?

ananiy:
@dbaeumer see my reply before, i am not good at english, sorry for my syntax...

ananiy:
> Which version of `tnpm` are you using and where can I find it?

you can using `npm i -g cnpm`!

ananiy:
@dbaeumer anything update? help plz🙏

ananiy:
https://github.com/cnpm/cnpm

dbaeumer:
Thanks for the pointer. PR welcome.

ananiy:
how to debug this, i don't know what's wrong...

dbaeumer:
[...]

PR #1275, opened 2021-06-03
feat: Support cnpm

ananiy:
fix: #1270

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -736,7 +736,7 @@ const languageId2DefaultExt: Map<string, string> = new Map([
 
 const languageId2ParserRegExp: Map<string, RegExp[]> = function createLanguageId2ParserRegExp() {
 	const result = new Map<string, RegExp[]>();
-	const typescript = /\/@typescript-eslint\/parser\//;
+	const typescript = /@typescript-eslint\/parser/;
 	const babelESLint = /\/babel-eslint\/lib\/index.js$/;
 	result.set('typescript', [typescript, babelESLint]);
 	result.set('typescriptreact', [typescript, babelESLint]);


dbaeumer:
We can't simply tweak the regular expression. We need to

- understand how cnpm creates these folders
- may be we need to have a different set of regulare expressions for the cnpm case.

I want to ensure that this doesn't break any other setup.

ananiy:
cnpm flat packages and using symbol link to reduce install time, the real package name is not like the normal.

I test it for npm and yarn, it works well.

and I check the code effects, I think it reasonable to fix this, the regexp may not be so strict...

Anyway, looking forward to fixing this issue soon, thank you very much!!!!!!!!!!

dbaeumer:
I am not saying it is not reasonable to fix this, but simply relaxing one regular expression since that is your use case is not the right thing to do. As said we need to handle this more general since such a simple fix might not be possible for other regular expressions (e.g. why did you only address this for typescript and not for babel)

ananiy:
when I install `babel-eslint`, I got: 
```bash
warning babel-eslint@10.1.0: babel-eslint is now @babel/eslint-parser. This package will no longer receive updates.
```

- babel-eslint
    - /Users/ananiy/demo/eslint-tnpm-demo/node_modules/babel-eslint/lib/index.js
    - /Users/ananiy/demo/eslint-tnpm-demo/node_modules/_babel-eslint@10.1.0@babel-eslint/lib/index.js
- @babel/eslint-parser
    - /Users/ananiy/demo/eslint-tnpm-demo/node_modules/@babel/eslint-parser/lib/index.cjs
    - /Users/ananiy/demo/eslint-tnpm-demo/node_modules/_@babel_eslint-parser@7.14.4@@babel/eslint-parser/lib/index.cjs
- @angular-eslint/template-parser
    - /Users/ananiy/demo/eslint-tnpm-demo/node_modules/@angular-eslint/template-parser/dist/index.js
    - /Users/ananiy/demo/eslint-tnpm-demo/node_modules/_@angular-eslint_template-parser@12.1.0@@angular-eslint/template-parser/dist/index.js

dbaeumer:
@ananiy thanks for the work. 

As I said before we shouldn't simply relax the regular expression to support `cnpm`. From my experience pattern matching should always be done using a precise pattern when possible. 

As I suggested earlier we should do the following:

- add `cnpm` to the list of supported package managers.
- check if `cnpm` needs any special command line to query for global install package location (e.g. the code here: https://github.com/microsoft/vscode-eslint/blob/main/server/src/eslintServer.ts#L695)
- make `languageId2ParserRegExp` package manager aware so that we can store different regexp depending on the package manager used (https://github.com/microsoft/vscode-eslint/blob/main/server/src/eslintServer.ts#L737)

ananiy:
I've updated the code and look forward to your suggestions! 
thanks a lot~

dbaeumer:
I will have a look end of the week. We have VS Code shipping week right now.

client/src/extension.ts
```diff
@@ -1841,7 +1843,7 @@ function realActivate(context: ExtensionContext): void {
 					`To use ESLint please install eslint by running ${localInstall[packageManager]} in the workspace folder ${workspaceFolder.name}`,
 					`or globally using '${globalInstall[packageManager]}'. You need to reopen the workspace after installing eslint.`,
 					'',
-					isPackageManagerNpm ? 'If you are using yarn or pnpm instead of npm set the setting `eslint.packageManager` to either `yarn` or `pnpm`' : null,
+					isPackageManagerNpm ? 'If you are using yarn or pnpm or cnpm instead of npm set the setting `eslint.packageManager` to either `yarn` or `pnpm` or `cnpm`' : null,
```
dbaeumer, line 1846: this must be `yarn, pnpm, or cnpm`

client/src/extension.ts
```diff
@@ -1861,7 +1863,7 @@ function realActivate(context: ExtensionContext): void {
 				client.info([
 					`Failed to load the ESLint library for the document ${uri.fsPath}`,
 					`To use ESLint for single JavaScript file install eslint globally using '${globalInstall[packageManager]}'.`,
-					isPackageManagerNpm ? 'If you are using yarn or pnpm instead of npm set the setting `eslint.packageManager` to either `yarn` or `pnpm`' : null,
+					isPackageManagerNpm ? 'If you are using yarn or pnpm or cnpm instead of npm set the setting `eslint.packageManager` to either `yarn` or `pnpm` or `cnpm`' : null,
```
dbaeumer, line 1866: same as above

server/src/eslintServer.ts
```diff
@@ -738,15 +745,33 @@ const languageId2ParserRegExp: Map<string, RegExp[]> = function createLanguageId
...
+	result.set('typescript', [typescript, babelESLint, newBabelEslint]);
+	result.set('typescriptreact', [typescript, babelESLint, newBabelEslint]);
 
 	const angular = /\/@angular-eslint\/template-parser\//;
 	result.set('html', [angular]);
 
 	return result;
 }();
 
+const cnpmLanguageId2ParserRegExp: Map<string, RegExp[]> = function() {
```
dbaeumer, line 759: to avoid false positives we should make the regular expression more cnpm specific (include there specific version flattening into the regular expression).

server/src/eslintServer.ts
```diff
@@ -738,15 +745,33 @@ const languageId2ParserRegExp: Map<string, RegExp[]> = function createLanguageId
...
+	result.set('typescript', [typescript, babelESLint, newBabelEslint]);
+	result.set('typescriptreact', [typescript, babelESLint, newBabelEslint]);
+
+	const angular = /@angular-eslint\/template-parser\//;
+	result.set('html', [angular]);
+
+	return result;
+}();
+
 const languageId2ParserOptions: Map<string, { regExps: RegExp[]; parsers: Set<string>; parserRegExps?: RegExp[] }> = function createLanguageId2ParserOptionsRegExp() {
```
dbaeumer, line 775: This should change into a hash table of hash tables. The first one is package manager, and the next one the regular expressions. We can have a default one for unknown managers.

dbaeumer (changes requested):

PR #1280, opened 2021-06-11
Update package.json

usernamehw:
Some of the items can be tweaked to look better in Settings UI:

- Links: use `markdownDescription` for it to use actual links instead of plain text
- Some settings had no description (They were added from README file)

PR made without building the extension.


dbaeumer:
Nice. Thanks for providing this!!

PR #1317, opened 2021-08-15
Supports ESLint v8.

ota-meshi:
This PR supports ESLint v8.
Specifically, change it to use the `ESLint` class instead of the removed `CLIEngine`.

However, for backward compatibility, if `CLIEngine` is present, use `CLIEngine`. See [`ESLintClassEmulator` class](https://github.com/microsoft/vscode-eslint/pull/1317/files#diff-f2161f943f61b946208678b9aa18cc18638bc3fe7cb19d2230ce939dc60165afR1496).

I checked that this change works by making sure that the marker appears in each of the environments where `eslint@8.0.0-beta.0`, `eslint@7.32.0`, and `eslint@6.8.0` are installed.

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -286,6 +286,12 @@ interface CLIOptions {
 	fix?: boolean;
 }
 
+interface ESLintClassOptions {
+	cwd?: string;
+	fixTypes?: string[];
+	fix?: boolean;
+}
+
 // { meta: { docs: [Object], schema: [Array] }, create: [Function: create] }
 interface RuleData {
 	meta?: {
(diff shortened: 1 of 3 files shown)


dbaeumer:
@ota-meshi thanks a lot for the PR. Will have a look this week. Saw that the first eslint 8.0.0 beta got released lately.

server/src/eslintServer.ts
```diff
@@ -899,7 +935,7 @@ function resolveSettings(document: TextDocument): Promise<TextDocumentSettings>
 					if (!settings.silent) {
 						connection.console.error(`Failed to load eslint library from ${libraryPath}. See output panel for more information.`);
 					}
-				} else if (library.CLIEngine === undefined) {
+				} else if (library.CLIEngine === undefined && library.ESLint === undefined) {
 					settings.validate = Validate.off;
 					connection.console.error(`The eslint library loaded from ${libraryPath} doesn\'t export a CLIEngine. You need at least eslint@1.0.0`);
```
dbaeumer, line 940: Should we also mention ESLint class. Something like `neither exports a CLIEngine nor an ESLint class`

dbaeumer (changes requested):
Great work. I would suggest we do the minor word tweak.

Then I will publish an new internal version of the plugin so that users can do some insider testing before we release a new version. Is this OK for you?

ota-meshi:
Thank you for the review! I changed the message.

>Then I will publish an new internal version of the plugin so that users can do some insider testing before we release a new version. Is this OK for you?

It's OK with me!

dbaeumer (approved):

PR #1331, opened 2021-09-03
Enable custom eslint module path

jcampalo:
This PR enables a new setting `eslintPath` to allow users to set a custom path where ESLint module is installed. Similar to other plugins approach [prettier-vscode](https://github.com/prettier/prettier-vscode#prettierprettierpath).

### DIFF
--- a/client/src/tasks.ts
+++ b/client/src/tasks.ts
@@ -36,15 +36,18 @@ class FolderTaskProvider {
 			return undefined;
 		}
 		try {
-			const command = await findEslint(rootPath);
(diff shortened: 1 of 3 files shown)


dbaeumer:
@jcampalo can you explain why you think this is necessary. There is already a setting that is honored and that is `eslint.nodePath` defined here https://github.dev/microsoft/vscode-eslint/blob/eeb9a38ee395c33d7f09ae96613d34238c22d78e/package.json#L82-L82`

jcampalo:
Hi @dbaeumer, you are right, that seems to work, I didn't get that it can be used for that purpose, thanks.

Issue #1357, opened 2021-10-19
eslint.rules.customizations add rule options


douglance:
Many eslint rules have additional configuration options that more finely tune how and when rules should be applied. Would it be possible to add those rules to the `eslint.rules.customizations` setting in VSCode? 

It would be nice to have a workflow where my local development environment can have rules that are more strict than what is required by the project I'm working in. There are mistakes that I make consistently that I can prevent using specific rules that my entire team might not agree would make sense to enforce.

edupsousa:
@dbaeumer may I take a chance on developing this feature? I made a first attempt at https://github.com/edupsousa/vscode-eslint/commit/f1d74b342f2813ebf0e1b20d9e272fe0d2e5ef68, but I will probably need some guidance on the best way to extend and parse configuration settings, writing documentation, and so on.

I tested this first attempt with ESLint version 8, and it looks like it's working fine. It reports and fixes using the override config, being the rule present or not in the original eslintrc. *. It also respects the settings on `eslint.codeActionsOnSave.rules` and `eslint.rules.customizations`.

If it's fine, I would like to open a draft pull request to keep working on and discussing this feature, ok?

@douglance I would like to point out that ESLint is moving in a different direction by deprecating the use of "personal" configuration (see https://github.com/eslint/rfcs/pull/32), maybe this could make sense in a "project" oriented view, but I really like the idea of keeping the IDE flexible and enforcing project rules through build and integration process.

dbaeumer:
@edupsousa can you please outline in detail how this would then look like in a settings file. What I don't want to do is to mirror all the possible ESLint rule settings in a VS Code settings file. We would need to maintain the config / JSON schema for it. In addition it would put us into a catchup game if things change.

edupsousa:
@dbaeumer I agree, mirroring the ESLint config is a bad idea (and exactly what I naively did at first). After thinking a little bit about this issue I realized that its already possible to accomplish the rule overriding through the `eslint.options` setting:

@douglance to use stricter rules on your local environment you can create a custom ESLint config file, and then on `.vscode/settings.json` you can override the default ESLint config using the option `overrideConfigFile` for ESLint version 8, or `configFile` for previous versions:

```json
{
  "eslint.options": {
    "overrideConfigFile": "custom-eslintrc.js"
  }
}
```

At first I thought that this option would complete ignore the ESLint config for the project, but looks like the override config file is merged on top of the project configuration.

Sample Gist:
[...]

Issue #1206, opened 2021-03-15
When `rulePath` is missing, error notification appears on every file edit


andrewbranch:
Every now and then, our custom ESLint rules directory goes missing for some reason, and fixing it would generally not be my absolute top priority that needs to interrupt everything else. Except, it is, because the extension makes it impossible to focus on anything else in VS Code until it’s taken care of:

<details>
<summary>Screen capture</summary>

![Kapture 2021-03-15 at 09 48 15](https://user-images.githubusercontent.com/3277153/111189701-999f0180-8573-11eb-9f3d-d0436e434757.gif)
</details>

dbaeumer:
The reason for this is that the ESLint server has no idea about the severity of such an error. What we can do is to add a 'Ignore for this session' button.

PR #1364, opened 2021-11-10
Minor fixes for eslint.codeActionsOnSave.rules mechanism

edupsousa:
Hi @dbaeumer,

I was studying a way to contribute with #1357 and stepped on two small issues: a bug leading to stale entries on `saveRuleConfigCache` object, and behavior different from the documentation when using `eslint.codeActionsOnSave.rules` with ESLint version 8.

The stale cache is being caused by the use of the file URI to delete cache entries when a file is closed, those entries are indexed by the file path, not the URI. This fix only gets the file path from the URI and uses it to delete the cache entries, it was the simplest solution, but maybe using the URI to index cache entries could be a better one?

The behavior problem is because the function `getSaveRuleConfig` always returns undefined if `eslint.useESLintClass == false`, this makes sense to ESLint version 7, but not for version 8 where the class is the only engine available. This fix removes the responsibility of checking whether the class is available or the `eslint.useESLintClass` from the function `getSaveRuleConfig`, and decides if the rules should not be processed by checking if `ESLintClass.isCLIEngine == true`.

Hope that helps,
Eduardo

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -553,9 +553,10 @@ function isOff(ruleId: string, matchers: string[]): boolean {
 	return true;
 }
 
-async function getSaveRuleConfig(filePath: string, settings: TextDocumentSettings  & { library: ESLintModule }): Promise<SaveRuleConfigItem | undefined> {
-	let result = saveRuleConfigCache.get(filePath);
-	if (result === null) {
+async function getSaveRuleConfig(uri: string, settings: TextDocumentSettings  & { library: ESLintModule }): Promise<SaveRuleConfigItem | undefined> {
+	const filePath = getFilePath(uri);
+	let result = saveRuleConfigCache.get(uri);
+	if (filePath === undefined || result === null) {
 		return undefined;
 	}
 	if (result !== undefined) {
@@ -586,10 +587,10 @@ async function getSaveRuleConfig(filePath: string, settings: TextDocumentSetting
 		return offRules.size > 0 ? { offRules, onRules } : undefined;
 	}, settings);
 	if (result === undefined || result === null) {
-		saveRuleConfigCache.set(filePath, null);
+		saveRuleConfigCache.set(uri, null);
 		return undefined;
 	} else {
-		saveRuleConfigCache.set(filePath, result);
+		saveRuleConfigCache.set(uri, result);
 		return result;
 	}
 }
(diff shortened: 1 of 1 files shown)


dbaeumer:
@edupsousa thanks for the PR. 

One change I would like to request: use also the URI for the `saveRuleConfigCache` to make it consistent with all the other caches. Otherwise this is always special.

server/src/eslintServer.ts
```diff
@@ -553,44 +553,44 @@ function isOff(ruleId: string, matchers: string[]): boolean {
 	return true;
 }
 
-async function getSaveRuleConfig(filePath: string, settings: TextDocumentSettings  & { library: ESLintModule }): Promise<SaveRuleConfigItem | undefined> {
-	let result = saveRuleConfigCache.get(filePath);
-	if (result === null) {
+async function getSaveRuleConfig(uri: string, settings: TextDocumentSettings  & { library: ESLintModule }): Promise<SaveRuleConfigItem | undefined> {
```
edupsousa, line 556: Replaced `filePath` argument by the `uri`, as it is need here to `get` and `set` items on `saveRuleConfigCache`. Preferred this way instead of adding the `uri` to the arguments list to keep the function signature concise.

server/src/eslintServer.ts
```diff
@@ -553,44 +553,44 @@ function isOff(ruleId: string, matchers: string[]): boolean {
 	return true;
 }
 
-async function getSaveRuleConfig(filePath: string, settings: TextDocumentSettings  & { library: ESLintModule }): Promise<SaveRuleConfigItem | undefined> {
-	let result = saveRuleConfigCache.get(filePath);
-	if (result === null) {
+async function getSaveRuleConfig(uri: string, settings: TextDocumentSettings  & { library: ESLintModule }): Promise<SaveRuleConfigItem | undefined> {
+	const filePath = getFilePath(uri);
+	let result = saveRuleConfigCache.get(uri);
+	if (filePath === undefined || result === null) {
```
edupsousa, line 559: `filePath` still required here to be used with `ESLintClass.calculateConfigForFile` method, so if the uri schema couldn't be resolved to a file path we `return undefined` in the same way of when there's no rules.

server/src/eslintServer.ts
```diff
@@ -2283,7 +2283,7 @@ async function computeAllFixes(identifier: VersionedTextDocumentIdentifier, mode
 		connection.tracer.log(`Computing all fixes took: ${Date.now() - start} ms.`);
 		return result;
 	} else {
-		const saveConfig = filePath !== undefined && mode === AllFixesMode.onSave ? await getSaveRuleConfig(filePath, settings) : undefined;
+		const saveConfig = filePath !== undefined && mode === AllFixesMode.onSave ? await getSaveRuleConfig(uri, settings) : undefined;
```
edupsousa, line 2286: This is the only call to `getSaveRuleConfig` function, so I changed the call to use the uri instead of the filePath.
As a personal decision I kept the `filePath !== undefined` condition, even if we are not using `filePath` in the call anymore. I thought that it could be more efficient by avoiding an asynchronous call if we already know its return value, but I'm not really sure.

edupsousa:
@dbaeumer did the request changes to keep `saveRuleConfigCache` consistent with the other caches, please see the previous comments on some coding decisions and let me know if it needs any other changes.

dbaeumer:
Looks good to me.

dbaeumer (approved):

edupsousa:
@dbaeumer I have some spare time and would love to continue contributing to the project, do you want to point me to any specific issue that needs help with coding?

dbaeumer:
@edupsousa great. If you want you can have a look at: https://github.com/microsoft/vscode-eslint/issues/1206

If you need help simply ping me in the issue

Issue #1238, opened 2021-04-13
Annoying messages from eslint


BobNobrain:
Sometimes eslint just spams me with irrelevant parsing error messages (because it is enabled "on type"). For example, in the following code:

```tsx
import React from 'react';

export default function Test() {
    React.useEffect(/* cursor goes here */)
}
```

i keep getting error messages with this stacktrace:

```
[Error - 4:15:39 PM] TypeError: Cannot read property 'type' of undefined
Occurred while linting <project>/.../test.tsx:4
    at visitCallExpression (<project>/node_modules/eslint-plugin-react-hooks/cjs/eslint-plugin-react-hooks.development.js:1734:24)
    at <project>/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
    at Object.emit (<project>/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (<project>/node_modules/eslint/lib/linter/node-event-generator.js:256:26)
    at NodeEventGenerator.applySelectors (<project>/node_modules/eslint/lib/linter/node-event-generator.js:285:22)
    at NodeEventGenerator.enterNode (<project>/node_modules/eslint/lib/linter/node-event-generator.js:299:14)
    at CodePathAnalyzer.enterNode (/<project>/node_modules/eslint/lib/linter/code-path-analysis/code-path-analyzer.js:711:23)
    at <project>/node_modules/eslint/lib/linter/linter.js:954:32

    at Array.forEach (<anonymous>)
```

Is there a way to somehow mute this messages?

<img width="1379" alt="image" src="https://user-images.githubusercontent.com/9580615/114560597-8451eb80-9c75-11eb-8f1a-f073ca8e1fe5.png">

dbaeumer:
Currently no. All I can think of having a setting for this that avoid the notification.

BobNobrain:
I think this would be a reasonable solution. When the setting is enabled, it just converts all these errors into some warning messages that appear in extension's output channel, but do not open this error popup. (I also wonder if there is some way to aggregate these popups into a single one, but that probably cannot be done from within the extension).

egucciar:
+1 for this, its a constant headache since i updated VSCode

mkvlrn:
Yes, PLEASE. I've tried finding a way to completely disable VSCode notifications (there isn't a way) because of that.

[...]

PR #1375, opened 2021-11-21
Allow user to ignore server errors.

### LINKED ISSUE #1206: When `rulePath` is missing, error notification appears on every file edit

edupsousa:
This PR resolves #1206 by allowing users to ignore server errors _by error message_ until the server is restarted.

The error dialog is not displayed, but the error is still shown on the output channel.

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -1772,7 +1772,7 @@ function tryHandleMissingModule(error: any, document: TextDocument, library: ESL
 	return undefined;
 }
 
-const ignoredErrors: string[] = [];
+const ignoredErrors: Set<string> = new Set();
 
 function showErrorMessage(error: any, document: TextDocument): Status {
 	const errorMessage = `ESLint: ${getMessage(error, document)}. Please see the 'ESLint' output channel for details.`;
@@ -1780,13 +1780,13 @@ function showErrorMessage(error: any, document: TextDocument): Status {
 		{ title: 'Open Output', id: 1},
 		{ title: 'Ignore for this Session', id: 2}
 	];
-	if (!ignoredErrors.includes(errorMessage)) {
+	if (!ignoredErrors.has(errorMessage)) {
 		void connection.window.showErrorMessage(errorMessage, ...actions).then((value) => {
 			if (value !== undefined) {
 				if (value.id === 1) {
 					connection.sendNotification(ShowOutputChannel.type);
 				} else if (value.id === 2) {
-					ignoredErrors.push(errorMessage);
+					ignoredErrors.add(errorMessage);
 				}
 			}
 		});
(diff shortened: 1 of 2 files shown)


edupsousa:
Hi @dbaeumer while working on this PR I noticed that the status bar color is not changing on error or warning status, it's on purpose? Maybe with the option to silent server errors we should highlight the status bar to show to the user that something is wrong and the file is not being parsed, what do you think?

Another option that I thought about adding with this (or another) PR is some commands to allow the user to silent or display **all** server errors for the session. Could this be a good option when the error message changes on repeated errors?

Maybe this could be helpful to alleviate other issues like #1238

dbaeumer:
@edupsousa sorry for the delay :-)

We should definitely update the color. It did in the past, but the code got somehow lost :-). It is here: https://github.com/microsoft/vscode-eslint/blob/7568b4ce8413e890e5f7d99b18845ccd00fd3d1a/client/src/extension.ts#L834

The `background` is never updated. 

I would actually not add any additional commands. The reasons are:
- I think not showing the same message over and over is a good default
- clicking on the ESLint status takes you to the output channel.

server/src/eslintServer.ts
```diff
@@ -1769,12 +1769,27 @@ function tryHandleMissingModule(error: any, document: TextDocument, library: ESL
 	return undefined;
 }
 
+const ignoredErrors: string[] = [];
```
dbaeumer, line 1772: This should be a set

dbaeumer (changes requested):

server/src/eslintServer.ts
edupsousa: Did the change as requested.

edupsousa:
> @edupsousa sorry for the delay :-)
> 
> We should definitely update the color. It did in the past, but the code got somehow lost :-). It is here:
> 
> https://github.com/microsoft/vscode-eslint/blob/7568b4ce8413e890e5f7d99b18845ccd00fd3d1a/client/src/extension.ts#L834
> 
> The `background` is never updated.
> 
> I would actually not add any additional commands. The reasons are:
> 
> * I think not showing the same message over and over is a good default
> * clicking on the ESLint status takes you to the output channel.

No problem with the delay. The year is ending but the work just increases. :-)

I was playing around another day and already did the background color update, but the branch where I did it is polluted with other things that I was playing. I will do some cleanup and tests before submitting it in a new PR.

dbaeumer:
@edupsousa I looked at the code again and did you do any chances to update the background color? It is not in the PR.

edupsousa:
> @edupsousa I looked at the code again and did you do any chances to update the background color? It is not in the PR.

@dbaeumer, I was planning to put it on a separate PR, sorry for the confusion. But, to keep it simple I just added the status bar color code to this same PR.

dbaeumer:
@edupsousa thanks. Will have a look

PR #1399, opened 2022-01-07
Document functions

hahnbeelee:
Used this [AI Doc Writer extension](https://marketplace.visualstudio.com/items?itemName=mintlify.document) to document some of the functions in both the client and server


dbaeumer:
@hahnbeelee since all these function are internal I think there is no extra value of adding JSDoc to them. I will therefore close the PR.

PR #1404, opened 2022-01-17
fix: typescript probe not work properly on cnpm

denghongcai:
fix probe detection on `cnpm install` scene

if project use `cnpm` to manage dependencies, it will use soft link to flat dep

image

use regex like this https://github.com/microsoft/vscode-eslint/blob/main/server/src/eslintServer.ts#L997 will fix it.

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -981,8 +981,8 @@ function getBlockComment(languageId: string): [string, string] {
 
 const languageId2ParserRegExp: Map<string, RegExp[]> = function createLanguageId2ParserRegExp() {
 	const result = new Map<string, RegExp[]>();
-	const typescript = /\/@typescript-eslint\/parser\//;
-	const babelESLint = /\/babel-eslint\/lib\/index.js$/;
+	const typescript = /@typescript-eslint\/parser\//;
+	const babelESLint = /babel-eslint\/lib\/index.js$/;
 	result.set('typescript', [typescript, babelESLint]);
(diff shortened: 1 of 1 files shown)


denghongcai:
cnpm website: https://cnpmjs.org/

denghongcai:
@dbaeumer give a glance?

dbaeumer:
@denghongcai thanks for the PR.

In general I am not a fan of relaxing regexp since they might now match also they shouldn't. Especially removing path separators might lead to unexpected results.

To make this correctly work for cnpm I think we need to do the following:

- add cnpm as a supported package manager (see package manager setting)
- have special regular expression for cnpm.

denghongcai:
> @denghongcai thanks for the PR.
> 
> In general I am not a fan of relaxing regexp since they might now match also they shouldn't. Especially removing path separators might lead to unexpected results.
> 
> To make this correctly work for cnpm I think we need to do the following:
> 
> * add cnpm as a supported package manager (see package manager setting)
> * have special regular expression for cnpm.

i use regex just like https://github.com/microsoft/vscode-eslint/blob/main/server/src/eslintServer.ts?rgh-link-date=2022-01-17T14%3A16%3A52Z#L997

as you suggest, what should i need to do next? make a seperate pr to add cnpm as supported package manager first?

dbaeumer:
@denghongcai the regexp you are pointing to is a parser option and not a path on disk. Parser options usually have no path separator in them that is why I don't test for them.

Yes, we should add cnpm as a package manager first.

Issue #1474, opened 2022-05-30
packageManager set to pnpm will not work


Dunqing:
I see the following error in eslint output
<img width="1154" alt="image" src="https://user-images.githubusercontent.com/29533304/170914075-f6390b03-fd38-4130-a461-5c43844302cd.png">

After some investigation, I found out why. when my workspace already has `pnpm` installed and typing `pnpm root -g` in terminal will receive `Run "pnpm setup" to create it automatically, or set the global-bin-dir setting, or the PNPM_HOME env variable. The global bin directory should be in the PATH.`.

When `pnpm` is not installed in my workspace, it works.

dbaeumer:
See my last comment in https://github.com/microsoft/vscode-eslint/pull/1475.

Closing.

PR #1475, opened 2022-05-30
fix: pnpm is already installed in the current workspace and does not work.

Dunqing:
close: #1474

### DIFF
--- a/server/src/eslint.ts
+++ b/server/src/eslint.ts
@@ -1075,7 +1075,12 @@ export namespace ESLint {
 			pnpm: {
 				cache: undefined,
 				get(): string {
-					const pnpmPath = execSync('pnpm root -g').toString().trim();
+					let pnpmPath: string;
+					try {
+						pnpmPath = execSync('pnpm root -g').toString().trim();
+					} catch {
+						pnpmPath = execSync('pnpm root').toString().trim();
+					}
 					return pnpmPath;
 				}
 			}


Dunqing:
> Should we not first try locally and then globally?

No, `pnpm root` will return an path, whether it exists locally or not

itpropro:
Any progress on this? I have the same error as mentioned in #1474 and this would fix it 😉

dbaeumer:
Sorry, has to wait until next week. Busy with other things.

server/src/eslint.ts
```diff
@@ -1075,7 +1075,12 @@ export namespace ESLint {
 			pnpm: {
 				cache: undefined,
 				get(): string {
-					const pnpmPath = execSync('pnpm root -g').toString().trim();
+					let pnpmPath: string;
+					try {
+						pnpmPath = execSync('pnpm root -g').toString().trim();
+					} catch {
+						pnpmPath = execSync('pnpm root').toString().trim();
```
ivangabriele, line 1082: I guess you enclosed that as a `catch` to avoid any behavior breaking change. But unless we want to make that a customizable settings to force global pnpm use instead of local one, I don't think this is necessary, is it?

I would simply remove the `-g` flag from the original source.

Dunqing: see https://github.com/microsoft/vscode-eslint/pull/1475#issuecomment-1141206558

ivangabriele: That's exactly what I'm saying ;). Since `pnpm root` will always return the path, there is no need to keep the `pnpm root -g` line.

Dunqing: > That's exactly what I'm saying ;). Since `pnpm root` will always return the path, there is no need to keep the `pnpm root -g` line.

When you don't have `pnpm` installed locally, executing `pnpm root` will return a non-existent path, which will cause a more serious error.

Unless there is a better way to fix this problem, we must use `pnpm root -g` as a preference.

JanJakes: @Dunqing What about checking if the local path exists and falling back to `pnpm root -g`? 🤔

dbaeumer:
Sorry for the long delay. 

Disclaimer: I have little knowledge about `pnpm`.

When you are talking about having `pnpm` install locally then you mean in a node_modules folder. Right?

If this is the case then the proposed PR will not work since `execSync` will not look into the `./node_modules/.bin` folder. 

If this is incorrect what does it mean to have `pnpm` installed locally?

server/src/eslint.ts
Dunqing: > @Dunqing What about checking if the local path exists and falling back to `pnpm root -g`? 🤔

Sounds good.

Dunqing:
> When you are talking about having `pnpm` install locally then you mean in a node_modules folder. Right?

Yes.

> If this is the case then the proposed PR will not work since execSync will not look into the ./node_modules/.bin folder.

I don't think it's related to `execSync` .

dbaeumer:
But how should this find `pnpm` then ?

```
pnpmPath = execSync('pnpm root').toString().trim()
```

Dunqing:
> ```
> pnpmPath = execSync('pnpm root').toString().trim()
> ```



This is correct, but we need to check if the path exists https://github.com/microsoft/vscode-eslint/pull/1475#discussion_r935303452. 

image

If it doesn't exist, we run `pnpm root -g` again .

Or you can work with the current PR approach.

dbaeumer:
May be I still misunderstand this. In the example you gave me `pnpm` seems to be installed both globally and locally. If I only install `pnpm` locally and run it from a terminal I get a `Command 'pnpm' not found` message.

So this will not work if pnpm isn't installed globally as well. Right? And could this lead to some version mismatch where the global `pnpm` version is different than the local one ?

Dunqing:
> May be I still misunderstand this. In the example you gave me `pnpm` seems to be installed both globally and locally.

I have only installed globally.

> If I only install `pnpm` locally and run it from a terminal I get a `Command 'pnpm' not found` message.

Yes, that's right. Because this PR is just to solve the issue of #1474.

> So this will not work if pnpm isn't installed globally as well. Right?

Yes, this is the expected behavior, or we can finally fallback to `npx pnpm `.

> And could this lead to some version mismatch where the global `pnpm` version is different than the local one ?

Yes, but there shouldn't be any problems.

We should actually use the local `pnpm` first, but at the moment i don't know if we can tell if the local `pnpm` is available.

dbaeumer:
There is actually a way to find this out. Have a look at: https://github.com/microsoft/vscode-eslint/blob/8eeb994b794cafa2a5182069a170a81a7dcc2abc/server/src/eslint.ts#L833

This is how I find the `eslint` installation and `File.resolve` considers the node resolution rules.

tiagobento:
Is it possible to add a configuration option for users to manually configure where the extension should look for the `eslint` bin? Something like `"eslint.installationPath": "node_modules/.pnpm/eslint@7.26.0/node_modules/eslint"`

dbaeumer:
This more or less exist with the `eslint.nodePath` setting. When looking for eslint that path is considered as well.

dbaeumer:
@bene-we 

I tried a clean `pnpm` install and running `pnpm root -g` give me this:

```
ERROR  Unable to find the global bin directory
Run "pnpm setup" to create it automatically, or set the global-bin-dir setting, or the PNPM_HOME env variable. The global bin directory should be in the PATH.
```

I think that is the reason why the call actually fails. 

The reason why the eslint server wants the global installation directory is to be able to find eslint if it is not installed locally. So removing the `-g` is actually not what I want :-). 

Can someone example to me why `pnpm` needs additional setup for a global module location?

dbaeumer:
After running `pnpm setup` `pnpm root -g` works as expected.

bene-we:
> I tried a clean `pnpm` install and running `pnpm root -g` give me this:
> 
> ```
> ERROR  Unable to find the global bin directory
> Run "pnpm setup" to create it automatically, or set the global-bin-dir setting, or the PNPM_HOME env variable. The global bin directory should be in the PATH.
> ```
> 
> I think that is the reason why the call actually fails.
> 

@dbaeumer experienced this as well, I wrote my comment after fixing it using `pnpm setup`. Yet the problem still exists even if `pnpm root -g` works in all terminals but in eslint extension.

Is there anything that could help find the root cause for it? Some logging for instance?

dbaeumer:
You could try `"eslint.trace.server": "verbose"`

If this doesn't help I can provide you with steps how to debug this if you are willing to do so.

bene-we:
@dbaeumer just enabled the tracing, and voilà magically it also works with `pnp root -g` in `eslintServer.js`. I even uninstalled and reinstalled the VS Code extension in order to have the original build files. Seems like VS Code or the Eslint extension just needed a restart (which I didn't try after I encountered the issue).

For anyone else: Please try to get `pnpm root -g` to work in all terminals and do a fresh restart of your computer, seems like the issue is fixed with that. If not, feel free to comment in here so I could try to debug further.

latin-1:
I'm sorry, but are there any updates?

dbaeumer:
I will close the PR for now. As it turned out `pnpm` needs an additional setup step to ensure `pnpm root -g` works without failure. But this is necessary for the extension to work correctly.

Please ping if someone has an idea to avoid the additional setup step.

PR #1495, opened 2022-07-04
Use client capabilities

MariaSolOs:
In order for the server to remain flexible and useful to many clients, the server must check the client's capabilities in the following scenarios:
1. Before including the `codeActionProvider` in its capabilities, it must ensure that the client signals code action literal support via the `textDocument.codeAction.codeActionLiteralSupport` property.
2. Before registering for the "workspace/didChangeConfiguration", the server must check that such dynamic registration is allowed.

This PR adds both checks.


server/src/eslintServer.ts
```diff
@@ -383,6 +385,7 @@ namespace CommandIds {
 connection.onInitialize((_params, _cancel, progress) => {
```
dbaeumer, line 385: Minor thing. If a parameter is used we remove the underscore in front. The underscore is a tagging mechanism in TS to avoid compiler errors in these cases.

dbaeumer (changes requested):
Simple rename from _params to params

server/src/eslintServer.ts
MariaSolOs: My bad. I was unfamiliar with this convention, in my TS projects I normally only use underscores for unused parameters. Fixing it!

dbaeumer (approved):

aeschli (approved):

Issue #1505, opened 2022-07-25
Bug: Disabling rule with VSCode context menu broken when there is description


ilyub:
Reproduction:

1.
Disable some rule in VSCode:
![image](https://user-images.githubusercontent.com/70687432/180875163-e0696b8f-5df7-471a-91c1-0e4603653a94.png)
```ts
// eslint-disable-next-line rule1
```

2.
Add comment
```ts
// eslint-disable-next-line rule1 -- This rule is disabled because ...
```

3.
Disable another rule:
```ts
// eslint-disable-next-line rule1 -- This rule is disabled because ..., rule2
```

MariaSolOs:
I think that we can close this issue now...?

dbaeumer:
Correct. Thanks for pointing out.

PR #1506, opened 2022-07-26
Check for -- when disabling rules

MariaSolOs:
Fixes the problem described in #1505.

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -658,16 +658,16 @@ messageQueue.registerRequest(CodeActionRequest.type, async (params) => {
 	}
 
 	function getDisableRuleEditInsertionIndex(line: string): number {
-		let lineIndex = line.indexOf('--');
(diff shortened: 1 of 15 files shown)


server/src/eslintServer.ts
```diff
@@ -657,29 +657,43 @@ messageQueue.registerRequest(CodeActionRequest.type, async (params) => {
 		return action;
 	}
 
+	function getDisableRuleEditInsertionIndex(line: string): number {
+		let lineIndex = line.indexOf('--');
```
dbaeumer, line 661: I would prefer to name this variable index or charIndex. Since it denotes the character inside a line and not a line.

dbaeumer (changes requested):
Otherwise, the PR looks great. Thanks for providing it.

dbaeumer (approved):

sandy081 (approved):
LGTM

PR #1522, opened 2022-09-08
Add Experimental support of Flat Config

### LINKED ISSUE #1518: any plan for support eslint new config file `eslint.config.js`?
new eslint 8.23.0 release. and cli can use new config file.

https://github.com/eslint/eslint/pull/16235

Is there any plan for support eslint new config file `eslint.config.js`? 👀

uhyo:
Fixes #1518 

This PR adds a support of [ESLint's new configuration format](https://eslint.org/docs/latest/user-guide/configuring/configuration-files-new).

This feature is enabled by setting `eslint.experimentalUseFlatConfig` to true.

Note that currently `FlatESLint` needs to be imported from a special endpoint `eslint/use-at-your-own-risk`. Therefore we should treat this feature as experimental.


uhyo:
@dbaeumer Hi, could you take a look when you have time? Support by VSCode's ESLint extension is a a huge step towards the spread of ESLint's new config system. (I'm not an ESLint member though 😅 )

dbaeumer:
@uhyo unfortunately this has to wait a little until next month. I am very busy with task I have to complete in other areas.

VladimirGrenaderov:
@dbaeumer may be someone else can review this PR and take care about release (as alpha/beta, doesn't matter)? It doesn't look so complicated, but really blocks adoption of trash-less ESLint config.

client/src/client.ts
```diff
@@ -394,6 +394,7 @@ export namespace ESLintClient {
 				synchronize: {
 					fileEvents: [
 						Workspace.createFileSystemWatcher('**/.eslintr{c.js,c.cjs,c.yaml,c.yml,c,c.json}'),
+						Workspace.createFileSystemWatcher('**/eslint.config.js'),
```
MariaSolOs, line 397: You will also need to add this file to this filter:
https://github.com/microsoft/vscode-eslint/blob/e345da07609583fb1f4df636f9363482e380929f/client/src/client.ts#L114

server/src/eslint.ts
```diff
@@ -259,23 +259,34 @@ export type ESLintModule =
...
 } | {
 	// 8.0 <= version.
 	ESLint: ESLintClassConstructor;
 	CLIEngine: undefined;
+	FlatESLint: undefined;
+} | {
+	// 8.21.0 <= version, experimental endpoint that supports Flat Config
+	ESLint: undefined;
+	CLIEngine: undefined;
+	FlatESLint: ESLintClassConstructor;
```
MariaSolOs, line 277: From the use of this type this seems to refer the exports of [ESLint main's entry point](https://github.com/eslint/eslint/blob/a682562458948f74a227be60a80e10e7a3753124/lib/api.js). If @dbaeumer can confirm that this was the original intention, then this last object is wrong, since the `FlatESLint` export is still not part of the public API. Moreover, I don't see why we would need to make `ESLint` undefined here? As of today ESLint is still exposing that class.

server/src/eslint.ts
```diff
@@ -947,13 +970,51 @@ export namespace ESLint {
 		return resultPromise;
 	}
 
+	function getESLintManifest(workingDirectory: DirectoryItem | undefined, configType: ConfigType | undefined) {
```
MariaSolOs, line 973: I truly dislike having these heuristics for figuring out the configuration type to use, since they might not match how ESLint discovers the configuration file to use and then we might end up with inconsistent linting experiences when using the CLI command versus using this extension.

I'm just giving my 2¢ here, but for preserving backwards compatibility and easing the maintenance of this extension while ESLint upgrades its API, I think we should follow a similar approach as the one used with the `withESLintClass` setting when moving from `CLIEngine` to `ESLint`. We could have an opt-in `withFlatConfig` setting that when set to true (and using the right ESLint version) will import the corresponding `FlatConfig` and use that API, else it will default to the current behavior until ESLint replaces the `ESLint` class with the flat one and then that becomes the new default. I'm aware that this would mean that if someone has an `eslint.config.js` file then they need an extra step of toggling the `withFlatConfig` setting to make this extension work but with an experimental API that's still changing, I think this is a reasonable sacrifice.

server/src/eslint.ts
```diff
@@ -724,14 +737,15 @@ export namespace ESLint {
...
-	const projectFolderIndicators: [string, boolean][] = [
-		[ 'package.json',  true ],
-		[ '.eslintignore', true],
-		[ '.eslintrc', false ],
-		[ '.eslintrc.json', false ],
-		[ '.eslintrc.js', false ],
-		[ '.eslintrc.yaml', false ],
-		[ '.eslintrc.yml', false ]
+	const projectFolderIndicators: [string, boolean, ConfigType | undefined][] = [
+		[ 'eslint.config.js', true, 'flat-config' ],
```
MariaSolOs, line 741: I would find an array of typed objects easier to understand here.

server/src/eslint.ts
```diff
@@ -947,13 +970,51 @@ export namespace ESLint {
...
+	}
+
+
 	export function newClass(library: ESLintModule, newOptions: ESLintClassOptions | CLIOptions, useESLintClass: boolean): ESLintClass {
 		if (ESLintModule.hasESLintClass(library) && useESLintClass) {
 			return new library.ESLint(newOptions);
 		}
+		if (ESLintModule.hasFlatESLintClass(library) && useESLintClass) {
+			return new library.FlatESLint(newOptions);
+		}
```
MariaSolOs, line 1011: Why have this if the if-block in line 1015 would already handle it?

MariaSolOs: Also it seems like you're using the `withESLintClass` setting in a way that doesn't align with the documented description of the setting.

uhyo:
@MariaSolOs Thank you for review! I recreated this PR with the config approach. This is indeed more reasonable; the clutter in the `ESLintModule` type is no longer there.

I named the new config `experimentalUseFlatConfig` because ESLint is going to merge the Flat Config support into the regular `ESLint` class, so this config will finally be unneeded after the experimental period.

Hope you can recheck this.

MariaSolOs:
@uhyo Thanks to you for the quick updates! Before you do any further changes though I think that @dbaeumer (being the main maintainer of this extension) should take a look, since I don't want you to spend your precious time and effort unnecessarily :)

dbaeumer:
The code changes actually look good to me. However, I would like to ask to clean up the playground you added and have one small example that uses a flat config.

dbaeumer (changes requested):
Clean up playground as pointed out in general comment.

uhyo:
@dbaeumer Thanks! I cleaned up the playground.

$shared/settings.ts
```diff
@@ -155,6 +155,7 @@ export type ConfigurationSettings = {
 	validate: Validate;
 	packageManager: PackageManagers;
 	useESLintClass: boolean;
+	experimentalUseFlatConfig: boolean;
```
MariaSolOs, line 158: To remain consistent with the `useESLintClass` setting, I would remove the `experimental` suffix here.

package.json
```diff
@@ -182,6 +181,12 @@
 					"default": false,
 					"description": "Since version 7 ESLint offers a new API call ESLint. Use it even if the old CLIEngine is available. From version 8 on forward on ESLint class is available."
 				},
+				"eslint.experimentalUseFlatConfig": {
+					"scope": "resource",
+					"type": "boolean",
+					"default": false,
+					"description": "Enable support of experimental Flat Config (aka eslint.config.js, supported by ESLint version 8.21 or later)."
```
MariaSolOs, line 188: As mentioned previously, I would remove the `experimental` here, since flat config will eventually become stable and we don't want to have separate settings for the `experimental` and `stable` APIs.

package.json
```diff
@@ -338,7 +343,8 @@
 							"items": {
 								"type": "string"
 							}
-						}, {
+						},
+						{
```
MariaSolOs, line 347: Not too important, but unless you did this on purpose, you might have a formatter making changes here.

package.json
```diff
@@ -464,29 +471,29 @@
...
-										"off"
-									],
-									"type": "string"
-								},
-								"rule": {
-									"type": "string"
-								}
+					"items": {
+						"properties": {
+							"severity": {
```
MariaSolOs, line 476: What's the change here? Did you replace tabs with spaces?

server/src/eslint.ts
```diff
@@ -724,14 +724,18 @@ export namespace ESLint {
...
-		[ 'package.json',  true ],
-		[ '.eslintignore', true],
-		[ '.eslintrc', false ],
-		[ '.eslintrc.json', false ],
-		[ '.eslintrc.js', false ],
-		[ '.eslintrc.yaml', false ],
-		[ '.eslintrc.yml', false ]
+	const projectFolderIndicators: {
+		fileName: string;
+		isRoot: boolean;
```
MariaSolOs, line 729: Thanks for following my suggestion! This looks much cleaner 💄

server/src/eslint.ts
```diff
@@ -825,31 +829,58 @@ export namespace ESLint {
 			if (moduleResolveWorkingDirectory === undefined && settings.workingDirectory !== undefined && !settings.workingDirectory['!cwd']) {
 				moduleResolveWorkingDirectory = settings.workingDirectory.directory;
 			}
+
+			// During Flat Config is considered experimental,
+			// we need to import FlatESLint from 'eslint/use-at-your-own-risk'.
+			// See: https://eslint.org/blog/2022/08/new-config-system-part-3/
+			const eslintPath = settings.experimentalUseFlatConfig ? 'eslint/use-at-your-own-risk' : 'eslint';
 			if (nodePath !== undefined) {
```
MariaSolOs, line 837: I would suggest first trying to check if the `'eslint'` entry point exports a `FlatESLint` object. If it doesn't and the setting is enabled, try the experimental path. That way we won't need to update this code when flat config becomes stable.

server/src/eslint.ts
```diff
@@ -825,31 +829,58 @@ export namespace ESLint {
...
+					if (settings.experimentalUseFlatConfig) {
+						const lib = loadNodeModule<{ FlatESLint?: ESLintClassConstructor }>(libraryPath);
+						if (lib === undefined) {
+							settings.validate = Validate.off;
+							if (!settings.silent) {
+								connection.console.error(`Failed to load eslint library from ${libraryPath}. See output panel for more information.`);
+							}
+						} else if (lib.FlatESLint === undefined) {
+							settings.validate = Validate.off;
+							connection.console.error(`The eslint library loaded from ${libraryPath} doesn\'t neither exports a FlatESLint class.`);
```
MariaSolOs, line 858: I think there's a typo in this error message.

server/src/eslint.ts
```diff
@@ -825,31 +829,58 @@ export namespace ESLint {
...
-						path2Library.set(libraryPath, library);
+						library = loadNodeModule(libraryPath);
+						if (library === undefined) {
+							settings.validate = Validate.off;
+							if (!settings.silent) {
+								connection.console.error(`Failed to load eslint library from ${libraryPath}. See output panel for more information.`);
+							}
+						} else if (library.CLIEngine === undefined && library.ESLint === undefined) {
+							settings.validate = Validate.off;
+							connection.console.error(`The eslint library loaded from ${libraryPath} doesn\'t neither exports a CLIEngine nor an ESLint class. You need at least eslint@1.0.0`);
```
MariaSolOs, line 878: `neither exports`?

MariaSolOs (changes requested):

$shared/settings.ts
uhyo: Sorry for confusing you, but I'd like to keep the name `experimental` because, unlike `useESLintClass`, this config will be unnecessary in the future when Flat Config is considered stable. At that time the regular `ESLint` class will have the Flat Config support [as described here](https://eslint.org/blog/2022/08/new-config-system-part-3/#:~:text=The%20FlatESLint%20class%20is%20intended%20only%20as%20a%20preview%20of%20functionality%3B%20once%20we%20switch%20over%20to%20flat%20config%20permanently%2C%20the%20current%20ESLint%20class%20will%20be%20deleted%20and%20FlatESLint%20will%20be%20renamed%20to%20ESLint.), and this config can be removed,

package.json
uhyo: Sorry, I revert the formatting changes.

server/src/eslint.ts
uhyo: As described in [this comment](https://github.com/microsoft/vscode-eslint/pull/1522#discussion_r990755707) the `'eslint'` entry point will never a `FlatESLint` object, but the name `ESLint` is kept when Flat Config is stablized.

When Flat Config becomes stable, what user needs to do is to disable the `experimentalUseFlatConfig` support.

server/src/eslint.ts
uhyo: Thanks, fixed 🙂
I also updated error messages so it helps users after Flat Config is stabilized.

$shared/settings.ts
MariaSolOs: I understand now. Thanks for explaining!

package.json
MariaSolOs: You can dismiss this comment now that you've explained why we're using the `experimental` name :)

server/src/eslint.ts
```diff
@@ -851,11 +851,11 @@ export namespace ESLint {
 						if (lib === undefined) {
 							settings.validate = Validate.off;
 							if (!settings.silent) {
-								connection.console.error(`Failed to load eslint library from ${libraryPath}. See output panel for more information.`);
+								connection.console.error(`Failed to load eslint library from ${libraryPath}. If you are using ESLint lower than v8.21, try upgrading ESLint. If you are using a fairly new ESLint version, try disabling 'experimentalUseFlatConfig' config. See output panel for more information.`);
```
MariaSolOs, line 854: Sorry for the nit, but I think a clearer message would be `"If you are using ESLint v8.21 or earlier, try upgrading it. For newer versions, try disabling the 'experimentalUseFlatConfig' setting. See the output panel for more information."`.

Error messages should be as clear as possible to help users encountering them :)

MariaSolOs: Other than that, everything else LGTM! Thanks a lot @uhyo for your patience and effort, I'm sure a lot of people will very much appreciate your contribution ❤️

uhyo: Fixed the error message! Thank you for patiently working with me.

dbaeumer (approved):

dbaeumer:
@MariaSolOs since you guided most of that coding can you please approve the changes as well. Then I will merge it into the main branch.

MariaSolOs (approved):

MariaSolOs:
> @MariaSolOs since you guided most of that coding can you please approve the changes as well. Then I will merge it into the main branch.

@dbaeumer @uhyo I tested the changes locally and everything LGTM. Let's ship this! 🚀

karlhorky:
Amazing, thanks for the work from everyone here! 🙌 

I take it from the milestone this will be available in the `2.2.7` release? (because the current release on [the marketplace](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) is `2.2.6`)

From semver I would have imagined that this would actually go into the `2.3.0` or even `3.0.0` release, but the `2.2.next` milestone that this is a part of seems to indicate that it would instead be `2.2.7`.

MariaSolOs:
@dbaeumer is probably going to be the one handling the release, and although I do agree with @karlhorky that `2.2.7` doesn't feel like the right semver version, I would suggest including this as part of a `2.3.0` release since we want to keep `3.0.0` for the pull diagnostics change coming up with #1534 (and in this way editors that use push diagnostics will still benefit from this flat config support).

dbaeumer:
`2.2.next` only denote the next release. Doesn't denote a concrete version number. The version number I always decide when I do the release.  It will be some wired version number increase due to the fact how the market-place handles insider versions. See https://github.com/microsoft/vscode-eslint#release-notes

karlhorky:
Ok thanks, I'm not sure I understand this explanation in this Release Notes docs section, so I did a PR to try to challenge my assumptions about it:

https://github.com/microsoft/vscode-eslint/pull/1540

PR #1528, opened 2022-09-26
Introduce option to prevent underlining multiple lines

aleclarson:
I rebased #463 in hopes that you'll reconsider merging this useful feature.

Your main concern seems to be consistency, but that's not a concern shared by ESLint maintainers:

> In [#8004](https://github.com/eslint/eslint/issues/8004), it was decided that ESLint should report large ranges if a node with a large range is reported by a rule, and **allow integrations to display a smaller report if desired**. The reasoning is that different integrations will have different needs with regard to report ranges (for example, an integration that posts comments on GitHub might want to know the entire reported range, whereas **an editor integration might only want to highlight a subset of a reported range**).
> [*Source*](https://github.com/eslint/eslint/issues/10424#issuecomment-394526121)

https://user-images.githubusercontent.com/1925840/192356428-aa9cf4a2-5870-4cf9-ac3a-0204fb635c32.mov

### DIFF
--- a/$shared/settings.ts
+++ b/$shared/settings.ts
@@ -163,6 +163,7 @@ export type ConfigurationSettings = {
 	options: ESLintOptions | undefined;
 	rulesCustomizations: RuleCustomization[];
 	run: RunValues;
+	singleLineUnderline: boolean;
 	nodePath: string | null;
 	workspaceFolder: WorkspaceFolder | undefined;
 	workingDirectory: ModeItem | DirectoryItem | undefined;
--- a/$shared/settings.ts
+++ b/$shared/settings.ts
@@ -163,7 +163,9 @@ export type ConfigurationSettings = {
 	options: ESLintOptions | undefined;
 	rulesCustomizations: RuleCustomization[];
 	run: RunValues;
-	singleLineUnderline: boolean;
+	problems: {
+		shortenToSingleLine: boolean;
+	};
 	nodePath: string | null;
 	workspaceFolder: WorkspaceFolder | undefined;
 	workingDirectory: ModeItem | DirectoryItem | undefined;
--- a/server/src/eslint.ts
+++ b/server/src/eslint.ts
@@ -11,7 +11,7 @@ import { execSync } from 'child_process';
 import { TextDocument } from 'vscode-languageserver-textdocument';
 import {
 	Diagnostic, DiagnosticSeverity, DiagnosticTag, ProposedFeatures, Range, TextEdit, Files, DocumentFilter, DocumentFormattingRegistrationOptions,
-	Disposable, DocumentFormattingRequest, TextDocuments
+	Disposable, DocumentFormattingRequest, TextDocuments, uinteger
 } from 'vscode-languageserver/node';
 import { URI } from 'vscode-uri';
 
@@ -571,7 +571,7 @@ namespace Diagnostics {
 				},
 				end: {
 					line: startLine,
-					character: Number.MAX_SAFE_INTEGER,
+					character: uinteger.MAX_VALUE,
 				}
 			});
 			endLine = startLine;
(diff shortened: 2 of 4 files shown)


dbaeumer:
I think the video is a good example why underlining the first line is not a good idea. The error is about the fact that `role="button"` implies some other property. Only underlining <div might not be very helpful.

aleclarson:
@dbaeumer That's a tradeoff that the user should be able to make on their own

server/src/eslint.ts
```diff
@@ -556,12 +556,16 @@ namespace Diagnostics {
...
 		const message = problem.message;
 		const startLine = typeof problem.line !== 'number' || Number.isNaN(problem.line) ? 0 : Math.max(0, problem.line - 1);
 		const startChar = typeof problem.column !== 'number' || Number.isNaN(problem.column) ? 0 : Math.max(0, problem.column - 1);
-		const endLine = typeof problem.endLine !== 'number' || Number.isNaN(problem.endLine) ? startLine : Math.max(0, problem.endLine - 1);
-		const endChar = typeof problem.endColumn !== 'number' || Number.isNaN(problem.endColumn) ? startChar : Math.max(0, problem.endColumn - 1);
+		let endLine = typeof problem.endLine !== 'number' || Number.isNaN(problem.endLine) ? startLine : Math.max(0, problem.endLine - 1);
+		let endChar = typeof problem.endColumn !== 'number' || Number.isNaN(problem.endColumn) ? startChar : Math.max(0, problem.endColumn - 1);
+		if (settings.singleLineUnderline && endLine !== startLine) {
+			endLine = startLine;
+			endChar = text.split('\n')[startLine].length;
```
dbaeumer, line 567: This splits always the whole document which we should avoid. A better way to do this is to use the `TextDocument.getText` with a range where the end character is a MAX_NUMBER;

$shared/settings.ts
```diff
@@ -163,6 +163,7 @@ export type ConfigurationSettings = {
 	options: ESLintOptions | undefined;
 	rulesCustomizations: RuleCustomization[];
 	run: RunValues;
+	singleLineUnderline: boolean;
```
dbaeumer, line 166: We should make it more clear what this means. It should may be named like this:

```ts
problems: {
    shortenToSingleLine: boolean;
}
```

dbaeumer (changes requested):

dbaeumer (changes requested):

server/src/eslint.ts
```diff
@@ -556,12 +557,26 @@ namespace Diagnostics {
...
+		let endChar = typeof problem.endColumn !== 'number' || Number.isNaN(problem.endColumn) ? startChar : Math.max(0, problem.endColumn - 1);
+		if (settings.problems.shortenToSingleLine && endLine !== startLine) {
+			const startLineText = document.getText({
+				start: {
+					line: startLine,
+					character: 0,
+				},
+				end: {
+					line: startLine,
+					character: Number.MAX_SAFE_INTEGER,
```
dbaeumer, line 574: This should be uinteger.MAX_VALUE do to the restrictions we have in LSP.

dbaeumer (changes requested):
See comment about uinteger.MAX_VALUE

server/src/eslint.ts
aleclarson: Oops, didn't know that was a thing :)

dbaeumer:
The changes look good to me. 

@aleclarson you need to accept the CLA (see https://github.com/microsoft/vscode-eslint/pull/1528#issuecomment-1258456366). I will not be able to merge this without you accepting it.

aleclarson:

dbaeumer (approved):

PR #1534, opened 2022-10-09
Add pull diagnostic support

MariaSolOs:
This PR updates the server and client to use the pull diagnostics model.

Before completing this, there are a couple of places where I'm unsure if a change is needed (I marked those as `TODO`s in the code):
- https://github.com/MariaSolOs/vscode-eslint/blob/d5e1d71d99e3d2c3bc51349a0cc942756783cc32/server/src/eslintServer.ts#L34-L50
- https://github.com/MariaSolOs/vscode-eslint/blob/d5e1d71d99e3d2c3bc51349a0cc942756783cc32/server/src/eslintServer.ts#L254-L261

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -297,7 +297,7 @@ connection.onDidChangeWatchedFiles(async (params) => {
 	// Simply revalidate all files.
 	environmentChanged(false);
 	RuleMetaData.clear();
-	ESLint.ErrorHandlers.clearNoConfigRepoerted();
+	ESLint.ErrorHandlers.clearNoConfigReported();
 	ESLint.ErrorHandlers.clearMissingModuleReported();
 
 	await Promise.all(params.changes.map(async (change) => {
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -10,7 +10,8 @@ import {
 	createConnection, Diagnostic, Range, TextDocuments, TextDocumentSyncKind, TextEdit, Command, WorkspaceChange,
 	VersionedTextDocumentIdentifier, DidChangeConfigurationNotification, DidChangeWorkspaceFoldersNotification,
 	CodeAction, CodeActionKind, Position, TextDocumentEdit, Message as LMessage, ResponseMessage as LResponseMessage,
-	uinteger, ServerCapabilities, NotebookDocuments, ProposedFeatures, ClientCapabilities, FullDocumentDiagnosticReport, TextDocumentIdentifier
+	uinteger, ServerCapabilities, NotebookDocuments, ProposedFeatures, ClientCapabilities, FullDocumentDiagnosticReport,
+	TextDocumentChangeEvent, DocumentDiagnosticParams, UnchangedDocumentDiagnosticReport
 } from 'vscode-languageserver/node';
 
 import { TextDocument } from 'vscode-languageserver-textdocument';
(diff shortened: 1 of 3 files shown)


rchl:
Am I correctly assuming that this would somewhat degrade the functionality or performance in editors that don't support pull diagnostics model since some code related to canceling validation requests is removed?

I know this server is made by Microsoft employee and you are working for Microsoft too but it is used by many editors outside of VSCode and it would be nice to consider that.

dbaeumer:
@rchl this is correct. However, keeping both implementations is not a good path forward either. I therefore suggest that we move the version number to 3.x. Then other editors can still use 2.x.

server/src/eslintServer.ts
```diff
@@ -34,6 +34,7 @@ import LanguageDefaults from './languageDefaults';
 
 // The connection to use. Code action requests get removed from the queue if
 // canceled.
+// TODO: Not sure if this changes with pull diagnostics.
```
dbaeumer, line 37: This needs to stay like this.

server/src/eslintServer.ts
```diff
@@ -428,38 +282,47 @@ connection.onInitialized(() => {
...
 
-messageQueue.registerNotification(DidChangeWorkspaceFoldersNotification.type, (_params) => {
-	environmentChanged();
-});
+connection.onDidChangeConfiguration(() => environmentChanged());
 
-async function validateSingle(document: TextDocument, publishDiagnostics: boolean = true): Promise<void> {
-	// We validate document in a queue but open / close documents directly. So we need to deal with the
+async function validateSingle(params: DocumentDiagnosticParams): Promise<FullDocumentDiagnosticReport | UnchangedDocumentDiagnosticReport> {
+	// TODO: Not sure if this comment still applies.
```
dbaeumer, line 296: This shouldn't happen anymore.

dbaeumer (commented):
Add some comments for the todos

dbaeumer:
and thanks a lot for doing this. Highly appreciated.

server/src/eslintServer.ts
```diff
@@ -57,7 +58,43 @@ const documents: TextDocuments<TextDocument> = new TextDocuments(TextDocument);
 // So all validating will work out of the box since normal document events will fire.
 const notebooks = new NotebookDocuments(documents);
 
-// This makes loading work in a plain NodeJS and a WebPacked environment
+class DiagnosticVersions {
```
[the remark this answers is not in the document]
MariaSolOs: My reasoning behind doing this is because not all editors offer [this kind of client configuration](https://github.com/microsoft/vscode-languageserver-node/blob/e20440f6f7f4dd4b74a40fab8f1ac163d2e98ece/client/src/common/diagnostic.ts#L114-L136) for indicating which events should trigger a pull. By having this we can ensure that diagnostics are updated based on the value of the `eslint.run` setting.

server/src/eslintServer.ts
```diff
@@ -152,223 +188,36 @@ function inferFilePath(documentOrUri: string | TextDocument | URI | undefined):
...
+function environmentChanged(refreshDiagnostics = true) {
 	ESLint.clearSettings();
 	RuleSeverities.clear();
 	SaveRuleConfigs.clear();
 	ESLint.clearFormatters();
 
-	for (const document of documents.all()) {
-		messageQueue.addNotificationMessage(ValidateNotification.type, document, document.version);
+	if (refreshDiagnostics) {
+		connection.languages.diagnostics.refresh();
```
rchl, line 220: Should not be called if client doesn't support pull.
Node libraries don't check client capabilities so this will trigger crash.

rchl: Well, unless the intention is to completely stop support non-pull model...
Not sure if that's @dbaeumer meant in another comment.

MariaSolOs: To agree with the LSP specs, I'll add the capability check before sending the request.

dbaeumer:
My proposal is to only support pull model from 3.x on forward. It allows us to remove quite some code and makes the overall processing easier on the server side.

dbaeumer:
Pull support got added.

PR #1553, opened 2022-11-14
Allow proper languageConfig even if languageId not set

mickaelistria:
on TextDocument.
The languageId setting on TextDocument may be null. In such case, infer it from the URI and filename.

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -133,7 +133,7 @@ function inferFilePath(documentOrUri: string | TextDocument | URI | undefined): 
 			if (filePath !== undefined) {
 				const textDocument = documents.get(uri.toString());
 				if (textDocument !== undefined) {
-					const extension = LanguageDefaults.getExtension(textDocument.languageId);
+					const extension = LanguageDefaults.getExtension(textDocument);
 					if (extension !== undefined) {
 						const extname = path.extname(filePath);
 						if (extname.length === 0 && filePath[0] === '.') {
@@ -679,8 +679,8 @@ messageQueue.registerRequest(CodeActionRequest.type, async (params) => {
 	}
 
 	function createDisableLineTextEdit(textDocument: TextDocument, editInfo: Problem, indentationText: string): TextEdit {
-		const lineComment = LanguageDefaults.getLineComment(textDocument.languageId);
-		const blockComment = LanguageDefaults.getBlockComment(textDocument.languageId);
+		const lineComment = LanguageDefaults.getLineComment(textDocument);
+		const blockComment = LanguageDefaults.getBlockComment(textDocument);
 
 		// If the concerned line is not the first line of the file
 		if (editInfo.line - 1 > 0) {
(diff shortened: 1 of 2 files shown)


dbaeumer:
@mickaelistria could you explain in which situations the languageId is null or not set since according to the protocol this should never happen.

mickaelistria:
In Eclipse LSP4E client, we do not set the language id unless we have one to declare (I didn't realize it was not conforming to the spec).
But the actual issue is that for a polyglot language server such as eslint, it seems strange to enforce the client to set a languaeg id in the client; I know VSCode has such option, but I don't think it's necessary in most other clients. So when I open a file in Ecipse IDE, it just assumes -without issue so far- that the language server will behave "at best" without an explicit languageId defined.
For ESLint or other ployglot language servers, then it means we'd have to duplicate some logic in the client that is available on server side already; to create a "languageId" for a given file. While it might be useful in some cases, it seems a bit overkill in most cases and such mapping could be placed only on language server.
Note that we have already existing "workarounds" for that in LSP4E, that I could possibly use.

carbondamien13 (approved):

dbaeumer:
Actualy, we added that to the protocol since clients usually do need to decide the language for a file. How would they otherwise decide on something like syntax coloring. And most clients have settings to map an extension to a mime type / language id. So IMO it does make perfect sense that client provide this information.

I am not a big fan of merging this PR since it might start a discussion on why the property is mandatory in the spec when the ESLint server works without it.

mickaelistria:
> Actualy, we added that to the protocol since clients usually do need to decide the language for a file.

Clients maybe; but usually if a language server is "appliable" to a file, it means the client has already decided it's compatible and the LS can process it. The LS has the smartness to decide for itself, the "language mapping" is more a client-side integration startegy than something that requires to be part of a protocol.

> So IMO it does make perfect sense that client provide this information.

The language id are not standardized by the spec. We can easily imagine some other LS using "js" or "ts" instead of "javascript" or "typescript". As the possible values are not standard, then it means they become additional responsibility and effort on the integration layer (specific to the client/LS couple, N*M). The question is why mandating this extra responsiblity/work/effort is the LSs can succeed without it?

> I am not a big fan of merging this PR since it might start a discussion on why the property is mandatory in the spec when the ESLint server works without it.

Exactly the track I'd like us to follow ;) As I mentioned above, I think this properly might be a design issue in the protocol and making it optional would allow LSP to better serve its goal.

dbaeumer:
The language ids are document in the specification: see https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentItem

So, in my opinion a client should be able to provide the language id in an open request.

sennhann29 (approved):

sennhann29 (approved):

dbaeumer:
I will close the PR since:

- the language ids are document in the spec
- I don't see why clients should not be able to provide the id, especially since a lot of clients allow mappings from file extensions to content types.

PR #1560, opened 2022-11-23
Improve support for unwritten buffers

mskelton:
When using the ESLint language server in Neovim, creating a new unwritten buffer will cause the language server to crash due to `fs.realpathSync.native` throwing an exception since the file does not exist.  This change simply ignores errors from `fs.realpathSync.native` falling back to the original URI for the file.

It appears that the way this is handled for VS Code is that the extension will not call `getFileSystemPath` if the `uri.scheme` is `untitled`.  Neovim doesn't share this concept of an `untitled` URI scheme, but the change I made seems to still work fine in VS Code.

Thoughts?

### DIFF
--- a/server/src/paths.ts
+++ b/server/src/paths.ts
@@ -85,11 +85,17 @@ export function getFileSystemPath(uri: URI): string {
 		result = result[0].toUpperCase() + result.substr(1);
 	}
 	if (process.platform === 'win32' || process.platform === 'darwin') {
-		const realpath = fs.realpathSync.native(result);
-		// Only use the real path if only the casing has changed.
-		if (realpath.toLowerCase() === result.toLowerCase()) {
-			result = realpath;
-		}
+    try {
+      const realpath = fs.realpathSync.native(result);
+      // Only use the real path if only the casing has changed.
+      if (realpath.toLowerCase() === result.toLowerCase()) {
+        result = realpath;
+      }
+    } catch {
+      // Silently ignore errors from `fs.realpathSync` to handle scenarios where
+      // the file being linted is not yet written to disk. This occurs in editors
+      // such as Neovim for non-written buffers.
+    }
 	}
 	return result;
 }


dbaeumer:
@mskelton what is the URI that Neovim provides in this case. It must not necessarily be `untitled` but it can't be `file` either since a lot of code in the ESLint server expects that `file` is something that exists on disk.

mskelton:
@dbaeumer The URI that Neovim provides uses the `file` scheme regardless of if the file is written to disk or not.

dbaeumer:
That could let to strange behavior since the path is used to determine the eslint config file to use. 

IMO catch the exception is not the right approach for this since there are other cases in the code that check for `untitled`. IMO we should allow the client to announce the untitled scheme and then make the server behave correctly and Neovim should choose the scheme of its liking for unsaved files.

mskelton:
So, I opened an issue in Neovim and got some responses from the Neovim maintainers.

> Is there actually a reason we need to do that, or it's just an accident? If the LSP protocol expects untitled:// for every never-saved buffer I don't see why Nvim lsp client couldn't do that...


> I think it would be worth creating an issue in the language-server-protocol repository and ask about clarification if clients should use the untitled scheme if a file doesn't exist on disk. To me it seems currently kinda undefined, or at least not explicit enough if they are supposed to do that.
> 
> There are a couple of language servers which currently don't work well when neovim sends a didOpen event for files which don't exist (like eclipse.jdt.ls), while others work well

Based on their responses, they seem open to adding support for a different URI scheme, it just sounds like perhaps there needs to be more standardization around that.  Seems related to https://github.com/microsoft/language-server-protocol/issues/1030#issuecomment-661869686?

dbaeumer:
@mskelton IMO we don't need to standardize the `untitled` scheme. This can be something a client can freely decide. The LSP protocol allows all kind of schemes. You can have a `my-location` scheme if you want. But what clients should not do is to use `file` for things that are not on disk.

Does that make sense?

mfussenegger:
>  we don't need to standardize the untitled scheme. This can be something a client can freely decide. The LSP protocol allows all kind of schemes. You can have a my-location scheme if you want. But what clients should not do is to use file for things that are not on disk.

I think the specification should at least state clearly that `file` cannot be used for unsaved files, and that another scheme that's up to the client must be used instead. And then it probably should also clarify how this affects the willSave/didSave/open requests and notifications.  

The specification also mentions this for didOpen:

> The document’s content is now managed by the client and the server must not try to read the document’s content using the document’s Uri. 

And vscode-eslint sets the capability for it. To me this seems as if it's not compliant?

mskelton:
@dbaeumer I haven't heard anything new on this.  I'm a little unsure of what I can do since the Neovim folks thought some more clear direction in the LSP specs would be helpful but it doesn't sound like that's necessarily desired by the LSP spec.  If it's all client dependent, it makes it challenging to make a server work well for all clients and as @mfussenegger pointed out there does seem to be some lack of full LSP spec compliance of vscode-eslint, furthering the challenge of knowing what the path forward here should be.

Thoughts?

mfussenegger:
Maybe to add some more data points to this:

- Some other language servers struggle as well if they receive a file URI that's not accessible on disk. (E.g. https://github.com/eclipse/eclipse.jdt.ls/issues/1815). My current interpretation of the specification is that this is a bug in the server due to the `didOpen` definition

- Some other language servers report a diagnostic error when functionality requires the file and is not available. For example https://github.com/python-lsp/python-lsp-server with the mypy plugin will have an error diagnostic for imports if the file is unavailable


- To implement some kind of `untitled` scheme for unsaved buffers in Neovim it would help to have assurance that servers can deal with that, by putting it in the specification. Otherwise we might up in a opposite situation where servers can't handle `untitled` and the maintainers tell the clients that it's not clear from the specification that they should handle it. To implement it we'd also need to know how this is supposed to interact with `didOpen/didSave`. If we report a `didOpen` with `untitled` and then the user saves the buffer out to disk, should this close the `untitled` and result in another `didOpen` with `file` scheme?

dbaeumer:
Forgot to answer:

> should this close the untitled and result in another didOpen with file scheme?

Yes, this is how name changes of the URI are supposed to be handled by the client (see https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didRename)

mfussenegger:
> a document with a file scheme can of course be unsaved, but it is assumed that the file does exist on disk.

This is currently not the case for Neovim, and the specification doesn't mention this constraint. Would it be an option to add that - to mention that any `file` URI must exist on disk if a `didOpen` notification is sent?

I'd still think that there should be a mention of something like `untitled`, otherwise how are server authors supposed to know that they should be dealing with scheme's != `file`, and how they should deal with them? As a client author I'd still assume it's _not_ an option to use it if it's left undefined.

mskelton:
Adding https://github.com/microsoft/vscode-eslint/pull/1580 for context which updates vscode-eslint to check that the URI scheme is not equal to file rather than equal to untitled.  This means what's left is:

1. Update the LSP spec with guidelines about checking for `scheme != 'file'`
2. Update the Neovim LSP implementation to use some non-file scheme for unwritten buffers.  Whether this is `untitled` or something else, the updated LSP spec would make the exact name less important.

I'm happy to help where I can if given some direction, I have a feeling the LSP spec portion is something I could do though I'm not sure about the process for suggesting modifications to the spec like this one.

mskelton:
Also, I'm going to close this PR as #1580 should resolve this issue once the LSP spec is updated and any editor implementations like Neovim that need to update the URI scheme to something new are updated.

PR #1563, opened 2022-12-03
Add “ignore untitled” setting

### LINKED ISSUE #1562: Add support to disable linting for unsaved files.
image

This file is created by https://github.com/tjx666/vscode-adobe-extension-devtools/blob/master/src/photoshop/descriptorInfoEditor.ts#L15

I find eslint lint this file but doesn't use the eslint config in my project root

MariaSolOs:
Fixes #1562

### DIFF
--- a/package.json
+++ b/package.json
@@ -301,7 +301,7 @@
 					"scope": "resource",
 					"type": "boolean",
 					"default": false,
-					"description": "If true, untitled files won't be validated ESLint."
+					"description": "If true, untitled files won't be validated by ESLint."
 				},
 				"eslint.probe": {
 					"scope": "resource",
--- a/package.json
+++ b/package.json
@@ -297,7 +297,7 @@
 					},
 					"description": "An array of language ids which should be validated by ESLint. If not installed ESLint will show an error."
 				},
-				"eslint.validate.ignoreUntitled": {
+				"eslint.ignoreUntitled": {
 					"scope": "resource",
 					"type": "boolean",
 					"default": false,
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -297,6 +297,12 @@
 					},
 					"description": "An array of language ids which should be validated by ESLint. If not installed ESLint will show an error."
 				},
+				"eslint.validate.ignoreUntitled": {
```
dbaeumer, line 300: VS Code doesn't support to have setting keys as subkeys of already existing keys (e.g you can't have `eslint.validate` as a key and the subkey `eslint.validate.ignoreUntitled`). This will cause some trouble. So we need to find either a different name or support some not syntax in the `validate` array. Something like

```
"eslint.validate": ["!untitled"]
```

But that might leave the question which positive files we should choose in this case. All?

Or we go with the setting `eslint.ignoreUntitled`.

I would actually go with the second.

dbaeumer (changes requested):

package.json
MariaSolOs: I see. And I agree, the second alternative seems clearer.

package.json
```diff
@@ -297,6 +297,12 @@
 					},
 					"description": "An array of language ids which should be validated by ESLint. If not installed ESLint will show an error."
 				},
+				"eslint.ignoreUntitled": {
+					"scope": "resource",
+					"type": "boolean",
+					"default": false,
+					"description": "If true, untitled files won't be validated ESLint."
```
dbaeumer, line 304: The description reads strange. Shouldn't it be: `If true, untitled files won't be validated.`

dbaeumer:
@MariaSolOs see my additional wording question.

package.json
MariaSolOs: Oops, I forgot a `by` before `ESLint`.

dbaeumer (approved):

PR #1588, opened 2023-01-24
Handle unused directive problems

### LINKED ISSUE #1585: Auto Fix on Save does not fix unused disable directives unless there are other lint problems
In my eslint config, I've got the following activated:
```
reportUnusedDisableDirectives: true,
```

This causes eslint to emit a warning for unused disable directives and it properly shows as a warning in vscode:
image

In my vscode settings I've got code actions on save enabled:
```
"editor.codeActionsOnSave": {
  "source.fixAll": true
},
```

However, upon save, this warning is not autofixed.  Additionally, there is no "quick fix" shown in vscode as there would be for other fixable errors.

This warning is autofixable at command line by eslint using the `--fix` flag.

I have found that if there are other errors in the file then upon save both the other error as well as the unused directive will be autofixed.  Additionally, if I choose the quick fix "Fix all auto-fixable problems" for the other error, it will automatically fix the unused directive.  It appears that automatic fixing will only not occur if this is the only problem in the file.

It's possible I'm missing a configuration somewhere that would allow this to be fixed and if that is the case would appreciate guidance on how to configure it.  If that is not the case then please consider a feature request to the extension to treat this as a fixable error.

Thanks!

MariaSolOs:
Fixes #1585.

### DIFF
--- a/server/src/eslint.ts
+++ b/server/src/eslint.ts
@@ -138,11 +138,9 @@ export namespace RuleMetaData {
 			if (toHandle.length === 0) {
 				return;
 			}
-			addUnusedDisableDirectivesMeta(toHandle);
 			rulesMetaData = typeof eslint.getRulesMetaForResults === 'function' ? eslint.getRulesMetaForResults(toHandle) : undefined;
 			toHandle.forEach(report => handled.add(report.filePath));
 		} else {
-			addUnusedDisableDirectivesMeta(reports);
 			rulesMetaData = typeof eslint.getRulesMetaForResults === 'function' ? eslint.getRulesMetaForResults(reports) : undefined;
 		}
 		if (rulesMetaData === undefined) {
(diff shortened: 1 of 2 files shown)


server/src/eslintServer.ts
```diff
@@ -845,7 +845,7 @@ messageQueue.registerRequest(CodeActionRequest.type, async (params) => {
 			});
 		}
 
-		if (settings.codeAction.disableRuleComment.enable) {
+		if (settings.codeAction.disableRuleComment.enable && ruleId !== RuleMetaData.unusedDisableDirectiveId) {
```
MariaSolOs, line 848: Since unused-directive problems are disabling comments themselves, we cannot disable them.

server/src/eslint.ts
```diff
@@ -157,15 +162,27 @@ export namespace RuleMetaData {
 	}
 
 	export function getUrl(ruleId: string): string | undefined {
+		if (ruleId === unusedDisableDirectiveId) {
```
dbaeumer, line 165: Instead of doing all these ifs would it make sense to add a fake RuleMetaData literal to the ruleId2Meta map.

dbaeumer:
@MariaSolOs please see my comment above.

server/src/eslint.ts
MariaSolOs: I thought about that, but (this is a very silly question) at which point would it make sense to add the fake rule? It isn't clear to me when this dictionary is created.

server/src/eslint.ts
```diff
@@ -167,6 +179,10 @@ export namespace RuleMetaData {
 	export function hasRuleId(ruleId: string): boolean {
 		return ruleId2Meta.has(ruleId);
 	}
+
+	export function isUnusedDisableDirectiveProblem(problem: ESLintProblem): boolean {
+		return problem.ruleId === null && problem.message.startsWith('Unused eslint-disable directive');
```
dbaeumer, line 184: One last question: I checked myself and I think the answer is no: but does ESLint support localization. If not right now then we are fine but this check might break in the future.

dbaeumer:
Thanks for the ping. See my last comment. I always wanted to ask that but I forgot about it.

server/src/eslint.ts
MariaSolOs: Good question. All I could find was [this proposal](https://github.com/eslint/eslint/issues/9870) but it looks stale, and for this specific diagnostic the message is hardcoded [here](https://github.com/eslint/eslint/blob/a59a4e6e9217b3cc503c0a702b9e3b02b20b980d/lib/linter/apply-disable-directives.js#L260-L261).

MariaSolOs: So I think we're good for now...

dbaeumer (approved):

aeschli (approved):

shawnmcknight:
Thanks @MariaSolOs !

PR #1591, opened 2023-01-26
Ensure nodeExit is always called

MariaSolOs:
Because sending the `ExitCalled` notification might fail, this is to ensure that `process.exit` always invokes its overridden version.


dbaeumer:
IMO we can't get rid of the timeout. The problem is that the underlying send is async and yields and if we don't wait a little it is very likely that the message never hits the wire since the process exists before.

This can of course even happen with the timeout but is far less likely. Thoughts?

PR #1611, opened 2023-02-17
Escape special characters in regular expressions for comments

### LINKED ISSUE #1610: Comment check regular expressions for quick actions use unescaped comment indicators
While using VS Code eslint plugin to check JSON files with [`eslint-plugin-jsonc`](https://github.com/ota-meshi/eslint-plugin-jsonc), I found out the quick actions to not work, even thought reporting and formatting worked just fine. In the output of the ESLint extension I could see the following error:

```
[Error - 20:46:34] Request textDocument/codeAction failed.
  Message: Request textDocument/codeAction failed with message: Invalid regular expression: //** eslint-disable-next-line/: Nothing to repeat
  Code: -32603 
```

The error comes from `createDisableLineTextEdit` here:

https://github.com/microsoft/vscode-eslint/blob/560df864580a9381d84b2f1e408f46f6600fda48/server/src/eslintServer.ts#L700

It tries to see if the previous line already has a eslint-disable comment. For that is composes a regular expression with the result of `getBlockComment` here:

https://github.com/microsoft/vscode-eslint/blob/560df864580a9381d84b2f1e408f46f6600fda48/server/src/languageDefaults.ts#L32

Because the language JSON is not in the `languageId2Config` in the same file. It falls back to `/**`. The resulting regular expession is thus:

```js
new RegExp(`/** eslint-disable-next-line`)
```

This is not a valid regular expression, since `/*` already means "as many `/` as you like" but then another `*` follows which has nothing to match on.

For most other languages the start of the block comment is defined as `/*`, which is actually a valid regular expression, but does not match the start of a block comment correctly. So if you happen to write code like this (highly unlikely, I know) the line with `foo` is recognized as comment line:

```tsx
    const foo = ' eslint-disable-next-line ';
    const bar: any = 'baz';
```

The quick action to add an eslint disable comment on top of the second line (e.g. for `no-explicit-any` or the unused variable) will just fail with the error:

```
[Error - 20:55:41] Failed to apply command: eslint.applyDisableLine
````

It would be cool to replace any special regex characters in the comment indicators when constructing the regular expression.

### PR DESCRIPTION
**Pull request by bxt**
Escape special characters in regular expressions for comments.

When figuring out if there is already an `eslint-disable` comment on the line before, the extension uses regular expressions to match comments. But comments include characters that appear naturally in regular expressions, like `*`. So the need to be replaced for construction of the regular expression. Otherwise adding the comment can fail when the previous line happens to include ` eslint-disable-next-line` even if not in a comment, or in languages that are not in the pre-defined list and thus use `/**` as a block comment indicator.

The regular expression to escape regular expressions is inspired by [ones commonly found e.g. on StackOverflow](https://stackoverflow.com/a/6969486/935676).

Fixes #1610 see there for more background and explanation.

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -680,6 +680,14 @@ messageQueue.registerRequest(CodeActionRequest.type, async (params) => {
 		return charIndex;
 	}
 
+	/**
+	 * Prefix characters with special meaning in comment markers with a backslash
+	 * See also: https://github.com/microsoft/vscode-eslint/issues/1610
+	 */
+	function escapeStringRegexp(value: string) {
+		return value.replace(/[|{}\\()[\]^$+*?.]/g, '\\$&');
+	}
+
 	function createDisableLineTextEdit(textDocument: TextDocument, editInfo: Problem, indentationText: string): TextEdit {
 		const lineComment = LanguageDefaults.getLineComment(textDocument.languageId);
 		const blockComment = LanguageDefaults.getBlockComment(textDocument.languageId);
@@ -691,13 +699,13 @@ messageQueue.registerRequest(CodeActionRequest.type, async (params) => {
 
 			// For consistency, we ignore the settings here and use the comment style from that
 			// specific line.
-			const matchedLineDisable = new RegExp(`${lineComment} eslint-disable-next-line`).test(prevLine);
+			const matchedLineDisable = new RegExp(`${escapeStringRegexp(lineComment)} eslint-disable-next-line`).test(prevLine);
 			if (matchedLineDisable) {
 				const insertionIndex = getDisableRuleEditInsertionIndex(prevLine, lineComment);
 				return TextEdit.insert(Position.create(editInfo.line - 2, insertionIndex), `, ${editInfo.ruleId}`);
 			}
 
-			const matchedBlockDisable = new RegExp(`${blockComment[0]} eslint-disable-next-line`).test(prevLine);
+			const matchedBlockDisable = new RegExp(`${escapeStringRegexp(blockComment[0])} eslint-disable-next-line`).test(prevLine);
 			if (matchedBlockDisable) {
 				const insertionIndex = getDisableRuleEditInsertionIndex(prevLine, blockComment);
 				return TextEdit.insert(Position.create(editInfo.line - 2, insertionIndex), `, ${editInfo.ruleId}`);

### CODE REVIEW

**Comment by bxt** [comment]

File: server/src/escapeStringRegexp.ts [hunk 1]
```diff
@@ -0,0 +1,3 @@
+export function escapeStringRegexp(string: string) {
```
dbaeumer, line 1: Can we move this into eslintServer since it is only used there.

dbaeumer (changes requested):

server/src/escapeStringRegexp.ts
bxt: @dbaeumer Sure, I [moved](https://github.com/microsoft/vscode-eslint/compare/a13fb0851d443ebdaddad759d8085d7ec3720982..ed602f24f0667b502ecd9a3bf209130aeb6d56d0) it close to where it's used now.

server/src/eslintServer.ts
```diff
@@ -680,6 +680,10 @@ messageQueue.registerRequest(CodeActionRequest.type, async (params) => {
 		return charIndex;
 	}
 
+	function escapeStringRegexp(string: string) {
```
MariaSolOs, line 683: I think that a comment referencing this PR or the issue would be helpful for other contributors to understand why this escaping is necessary.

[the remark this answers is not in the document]
dbaeumer: Agree with @MariaSolOs here. I usually call this `value` in these cases.

dbaeumer (changes requested):

dbaeumer (changes requested):
See the comment about renaming `string` to `value`

wesleybl:
@bxt could you please make the requested corrections so we can merge. It would be great to have this fix.

bxt:
Hey @dbaeumer @MariaSolOs I did the changes as you suggested, renamed the parameter `string` to `value` and added an explanatory comment with a link to the issue. Could you have another look?

MariaSolOs (approved):

vrubezhny:
@dbaeumer Any chance to see this PR merged?

dbaeumer (approved):

bxt:
it seems for the "Community PR Approvals" to become green I need one more review, I'm not sure who is eligible, but I see @aeschli is doing reviews here, maybe you could have look as well and review/merge this?

ptrxyz:
Yep, a single more review for these few lines and we would be back on track. I find @joaomoreno also doing some reviews, can we get a quick peek from either of you? :)

aeschli (approved):

AndreMaz:
@bxt @aeschli @dbaeumer I'm getting 

```bash
[Info  - 1:57:00 PM] ESLint server is starting.
[Info  - 1:57:01 PM] ESLint server running in node v20.9.0
[Info  - 1:57:01 PM] ESLint server is running.
[Info  - 1:48:05 PM] ESLint library loaded from: /home/Dev/web/node_modules/eslint/lib/api.js
[Info  - 1:48:05 PM] Server process exited successfully
[Error - 1:55:18 PM] Request textDocument/codeAction failed.
  Message: Request textDocument/codeAction failed with message: Range#create called with invalid arguments[[object Object], [object Object], undefined, undefined]
  Code: -32603 
  ````

when I create new file in a [t3-turbo](https://github.com/t3-oss/create-t3-turbo) based template.

Running `ESLint: Restart ESLint Server` solves the issue, ie, the I see no error and linting works as expected.


Context:
Plugin
```bash
ESLint: v3.0.10
```

vscode:
```bash
Version: 1.91.0
Commit: ea1445cc7016315d0f5728f8e8b12a45dc0a7286
Date: 2024-07-01T18:52:41.959Z
Electron: 29.4.0
ElectronBuildId: 9728852
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Linux x64 5.15.0-113-generic
```

dbaeumer:
@AndreMaz any steps on how to reproduce this. Otherwise it is hard to tell why it is failing.

soulchild:
> when I create new file in a [t3-turbo](https://github.com/t3-oss/create-t3-turbo) based template.
> Running `ESLint: Restart ESLint Server` solves the issue, ie, I see no error and linting works as expected.

Seeing the exact same thing as well. Every time I create a new file in my create-t3-turbo repo, all hell breaks lose and the ESLint server crashes with that error message. A restart via the command palette fixes the issue. Did you find a root cause and solution for this?

AndreMaz:
Hey @soulchild 
> Did you find a root cause and solution for this?

No, I got re-assigned to different stuff and didn't had time to investigate. 

Would you mind creating a repro repo to make this thing easier to debug?

Issue #1281, opened 2021-06-11
Toggling ESLint on and off for the workspace


skanev:
Hey.

Previous ESLint could be toggled on and off for the workspace. I found this useful, as warnings are incredibly distracting some of the time I'm writing code. To explain what I mean:

1. I decide to introduce a new function
2. I go and define it before having any callers
3. I'll add a caller in a bit, but I want to fiddle with it for a while and make sure it has the right signature
4. While this is going on, I'm keenly interested in (a) TypeScript errors, and uninterested in (b) ESLint errors. I'm getting a bunch of stuff highlighted red, however, and I have to go and check to see if it's (a) or (b), because I need to action (a) now and I'd postpone actioning (b) until I'm ready. Being able not get ESLint errors at this stage is valuable for me
5. Once I'm done and ready to test, I'd turn eslint on again and see if there is an error

In earlier version of the plugin, there was a Toggle ESLint command. I can't find it anymore. Am I missing something? If not, can it be reimplemented?

dbaeumer:
Since VS Code now offers to enabled/disabled extensions per workspace I remove the UI around it  from the ESLint extension. This avoids shipping a duplicate feature set.

Please ping if this doesn't work for you.

skanev:
@dbaeumer It doesn’t.

I have to reload Code to do it and there is no command ao it’s very involved.

also, I don’t need it strictly  per workspace - per file works to.

dbaeumer:
I only removed the commands (UI), the setting is still there `eslint.enable` though deprecated. Can you let me now if the setting is in general all you need?

skanev:
Almost.

Changing it stops new problems from appearing (which is what I want), but doesn't clear existing problems (which is also what I want). If you can de-deprecate (reprecate? heh) the setting and add back the Toggle command, to have it also clear problems, it would be perfect.

fritzfr:
I'm also in favour of bringing this back. I always used it in the past for when I just want to read code and not have every single line underlined in red. Please bring this back.

dbaeumer:
PR welcome :-)

dbaeumer:
Version 2.2.0 has the setting reenabled.

PR #1614, opened 2023-03-06
Reintroduce the enable and disable commands

kunev:
This addresses #1281. It reintroduces the `Enable ESLint` and `Disable ESLint` commands that were removed in c54770579ed6f77cc278f799fd807cf6816c9b56.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -92,6 +92,52 @@ export function activate(context: ExtensionContext) {
 		}
 	}
 
+	function enable() {
+		const folders = Workspace.workspaceFolders;
+		if (!folders) {
+			void Window.showWarningMessage('ESLint can only be enabled if VS Code is opened on a workspace folder.');
+			return;
+		}
+		const disabledFolders = folders.filter(folder => !Workspace.getConfiguration('eslint', folder.uri).get('enable', true));
+		if (disabledFolders.length === 0) {
+			if (folders.length === 1) {
+				void Window.showInformationMessage('ESLint is already enabled in the workspace.');
+			} else {
+				void Window.showInformationMessage('ESLint is already enabled on all workspace folders.');
+			}
+			return;
+		}
+		void pickFolder(disabledFolders, 'Select a workspace folder to enable ESLint for').then(folder => {
+			if (!folder) {
+				return;
+			}
+			void Workspace.getConfiguration('eslint', folder.uri).update('enable', true);
+		});
+	}
+
+	function disable() {
+		const folders = Workspace.workspaceFolders;
+		if (!folders) {
+			void Window.showErrorMessage('ESLint can only be disabled if VS Code is opened on a workspace folder.');
+			return;
+		}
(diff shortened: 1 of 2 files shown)


kunev:
@microsoft-github-policy-service agree company="Dext"

dbaeumer:
Actually I am not a fan of adding these again. The recommended way in VS Code is to use the enablement in the extensions viewlet. If that feature had existed before I started with VS Code ESLint I would have never added the setting. In addition the new profiles even allow you to enable extension on a profile basis. 

Why would you need to have these back.

kunev:
Same reason described in the linked issue above: It's convenient while writing completely new code to focus only on TS errors and then once everything is working reenable eslint to address what it has to say.

Doing this through the extensions viewlet requires opening it, searching for easlint, clicking on the button and then reloading vscode. Being able to achieve the same thing from the command palette is much more convenient. Additionally that way you could also assign a keyboard shortcut to these commands.

MariaSolOs:
@kunev I agree with @dbaeumer, it doesn't feel right to have a custom implementation for something that is already supported by the editor. Also remember that reloading VS Code can be done with the `Developer: Reload window` command, and the extensions tab has a search bar, so disabling/enabling the extension shouldn't even take a minute.

I still understand your point though, and how this can be disruptive in your development workflow. Maybe these are suitable alternatives:
- Create an `npm` script to run `eslint` before building or packing your project, so that you don't get those linting errors during development (or configure a linting step in your CI if applicable). There are also other products to add "hooks" for performing tasks before each `git` commit.
- Use the `Lint workspace` command to explicitly indicate when you want ESLint output.

kunev:
Those suggestions don't quite cover the same use case. However if you're not ok with adding these back there are plugins that allow toggling arbitrary settings in this manner, so those would be a viable alternative. That would just rely on the `eslint.enabled` setting being available.

zentus:
@kunev I needed this functionality as well, so I created [a small extension](https://marketplace.visualstudio.com/items?itemName=zentus.toggle-eslint) for it.

It registers a command that toggles the Workspace Setting `eslint.enable` between `true` and `false`.

To use it, install the extension and type `Toggle ESLint` into the command palette.

To bind it to a key, add it to `keybindings.json` like so:

```js
{
    "key": "<key-here>",
    "command": "toggle-eslint.toggle"
}
```

The code may be outdated, I'm not sure, but it currently works for me in VSCode `1.76.0`.

If you find a better way to achieve this functionality, please let me know.

[Extension](https://marketplace.visualstudio.com/items?itemName=zentus.toggle-eslint)
[Source](https://github.com/zentus/vscode-eslint-toggle)

kunev:
There are several general purpose settings toggling plugins:
- [Setting Toggle](https://marketplace.visualstudio.com/items?itemName=Ho-Wan.setting-toggle)
- [Settings Toggle](https://marketplace.visualstudio.com/items?itemName=z0al.settings-toggle)
- [Toggle](https://marketplace.visualstudio.com/items?itemName=rebornix.toggle)

I'm sure there are a bunch more in the marketplace. Each of these can be used to achieve what this PR was aiming. Some are more powerful than others, depending on your needs.

I'm closing this.

PR #1646, opened 2023-04-28
Automatically detect packageManager

wilkinson4:
Deprecate the packageManager setting, and automatically detect it with the vscode command.


wilkinson4:
> @microsoft-github-policy-service agree

wilkinson4:
First time making a PR to a language server so please let me know if I missed something.

server/src/eslint.ts
```diff
@@ -824,7 +823,6 @@ export namespace ESLint {
 			if (settings.validate === Validate.off) {
 				return settings;
 			}
-			settings.resolvedGlobalPackageManagerPath = GlobalPaths.get(settings.packageManager);
```
dbaeumer, line 827: This code can't be removed since the property is referenced here: https://github.com/microsoft/vscode-eslint/blob/d8214448aa1ba338c84184239bfee66cea24fbaa/server/src/eslint.ts#L875 and the value is essential for resolving global node modules.

dbaeumer (changes requested):
See comment about resolvedGlobalPackageManagerPath

server/src/paths.ts
```diff
@@ -86,11 +86,11 @@ export function getFileSystemPath(uri: URI): string {
...
 	if (process.platform === 'win32' || process.platform === 'darwin') {
 		try {
-      const realpath = fs.realpathSync.native(result);
-      // Only use the real path if only the casing has changed.
-      if (realpath.toLowerCase() === result.toLowerCase()) {
-        result = realpath;
-      }
+			const realpath = fs.realpathSync.native(result);
+			// Only use the real path if only the casing has changed.
+			if (realpath.toLowerCase() === result.toLowerCase()) {
```
MariaSolOs, line 91: It seems like your PR made some unintended whitespace changes. I suggest removing them for a cleaner `git` history.

client/src/client.ts
```diff
@@ -232,13 +232,15 @@ export namespace ESLintClient {
...
-		client.onRequest(NoESLintLibraryRequest.type, (params) => {
+		client.onRequest(NoESLintLibraryRequest.type, async (params) => {
 			const key = 'noESLintMessageShown';
 			const state = context.globalState.get<NoESLintState>(key, {});
 
 			const uri: Uri = Uri.parse(params.source.uri);
 			const workspaceFolder = Workspace.getWorkspaceFolder(uri);
-			const packageManager = Workspace.getConfiguration('eslint', uri).get('packageManager', 'npm');
+			const packageManager = await commands.executeCommand<'npm' | 'pnpm' | 'yarn'>(
+				'npm.packageManager'
```
MariaSolOs, line 242: You can use the [shared type](https://github.com/microsoft/vscode-eslint/blob/28910c7e2310d7a236198d47b269829066d4fdc7/%24shared/settings.ts#L150) instead of redefining it again here.

README.md
```diff
@@ -187,7 +187,6 @@ This extension contributes the following variables to the [settings](https://cod
     "eslint.lintTask.options": "-c C:/mydirectory/.eslintrc.json --ignore-path C:/mydirectory/.eslintignore ."
   }
   ```
-- `eslint.packageManager`: controls the package manager to be used to resolve the ESLint library. This has only an influence if the ESLint library is resolved globally. Valid values are `"npm"` or `"yarn"` or `"pnpm"`.
```
MariaSolOs, line 190: I don't think we should remove this. Instead we should mention that the setting is now deprecated and why it isn't needed anymore. This matches what we did with `eslint.autoFixOnSave`.

MariaSolOs (changes requested):

server/src/paths.ts
wilkinson4: These changes came from the lint script that I ran with the --fix option. I can undo them if we don't want the changes.

MariaSolOs: Oh, in that case yeah we should leave them to make sure the codebase is properly linted. Kind of inconvenient they're not in their original PRs though. Maybe we should have a lint PR check @dbaeumer?

MariaSolOs (approved):

server/src/eslint.ts
Skylerspencer: Thanks

wilkinson4:
@dbaeumer anything else that needs to be addressed in this PR?

dbaeumer:
IMO the change can be breaking for users that had different values set for ESLint and the package manager extension. IMO we should deprecated the value and code wise do the following:

- read ESLint's value AND the value through the command and consider their default values.

If both values are the same pint a message to the channel that the eslint value is deprecated and will be removed in the future. Use either of them

If the values are different print a message to the channel that the value is deprecated and that we will not honor it in the future and that we moved to `npm.packagemanager`. I would still use the eslint value for a while.

Breaking changes usually cause swirl especially since we will receive issue reports for it. This is why I try to avoid this. 

Sorry for detecting this so late.

Skylerspencer:
> Deprecate the packageManager setting, and automatically detect it with the vscode command.

no

client/src/client.ts
```diff
@@ -582,6 +582,18 @@ export namespace ESLintClient {
 			return clientOptions;
 		}
 
+		async function getPackageManager(uri: Uri) {
+			const userProvidedPackageManager:PackageManagers = Workspace.getConfiguration('eslint', uri).get('packageManager', 'npm');
+			const detectedPackageMananger = await commands.executeCommand<PackageManagers>('npm.packageManager');
+
+			if (userProvidedPackageManager === detectedPackageMananger) {
+				client.warn('The ESLint setting packageManager is deprecated, and will be removed in the future. Please update your configuration.', {}, true);
```
MariaSolOs, line 590: I think it would be useful to mention whether the provided manager matches the detected one, so that the user can act accordingly.

E.g.: In here log `"Detected package manager matches the one in the deprecated packageManager setting. This setting will be removed in the future, so you can remove it from your configuration."`, and in line 593 `"Detected package manager differs from the one in the deprecated packageManager setting. We will honor this setting until it is removed."`

MariaSolOs (approved):

client/src/client.ts
```diff
@@ -582,6 +582,18 @@ export namespace ESLintClient {
...
 
+		async function getPackageManager(uri: Uri) {
+			const userProvidedPackageManager:PackageManagers = Workspace.getConfiguration('eslint', uri).get('packageManager', 'npm');
+			const detectedPackageMananger = await commands.executeCommand<PackageManagers>('npm.packageManager');
+
+			if (userProvidedPackageManager === detectedPackageMananger) {
+				client.warn('Detected package manager matches the one in the deprecated packageManager setting. This setting will be removed in the future, so you can remove it from your configuration.', {}, true);
+				return detectedPackageMananger;
+			}
+			client.warn('Detected package manager differs from the one in the deprecated packageManager setting. We will honor this setting until it is removed.', {}, true);
```
dbaeumer, line 593: We should include the two values in the message so that the user already knows about them and can make a decision how to prceed.

client/src/client.ts
```diff
@@ -582,6 +582,18 @@ export namespace ESLintClient {
 			return clientOptions;
 		}
 
+		async function getPackageManager(uri: Uri) {
+			const userProvidedPackageManager:PackageManagers = Workspace.getConfiguration('eslint', uri).get('packageManager', 'npm');
+			const detectedPackageMananger = await commands.executeCommand<PackageManagers>('npm.packageManager');
+
+			if (userProvidedPackageManager === detectedPackageMananger) {
+				client.warn('Detected package manager matches the one in the deprecated packageManager setting. This setting will be removed in the future, so you can remove it from your configuration.', {}, true);
```
dbaeumer, line 590: I would actually not show a notification in this case (true -> false) since there is no action the user needs to take. Too many notification can be annoying.

dbaeumer (changes requested):
See other commants.

dbaeumer:
@wilkinson4 Forgot to say thanks for working on this.

wilkinson4:
> @wilkinson4 Forgot to say thanks for working on this.

No problem!

wilkinson4:
> See other commants.


@dbaeumer Following up on this one. Made the recommended changes.

dbaeumer (approved):

aeschli (approved):

PR #1651, opened 2023-05-11
Prepare Pre-release 2.4.1

dbaeumer:


aeschli (approved):

vrubezhny:
@dbaeumer Any reason why vscode-eslint/server version is not bumped for pre-release/2.4.1? 

https://github.com/microsoft/vscode-eslint/blob/main/server/package.json#L3

dbaeumer:
Since we don't publish the server separately. But I will increase the version.

PR #1691, opened 2023-07-27
fix: read only if `eslint.config.js` exists

kecrily:
Now if `eslint.experimental.useFlatConfig` is enabled but `eslint.config.js` is not actually present, then the legacy config file will not take effect even if it is present.

This makes it necessary to change the enabled state of `eslint.experimental.useFlatConfig` depending on the configuration format used by the project.

### DIFF
--- a/server/src/eslint.ts
+++ b/server/src/eslint.ts
@@ -866,10 +866,12 @@ export namespace ESLint {
 				moduleResolveWorkingDirectory = settings.workingDirectory.directory;
 			}
 
+			const ifUseFlatConfig = settings.experimental.useFlatConfig && fs.existsSync('eslint.config.js')
+
 			// During Flat Config is considered experimental,
 			// we need to import FlatESLint from 'eslint/use-at-your-own-risk'.
 			// See: https://eslint.org/blog/2022/08/new-config-system-part-3/
-			const eslintPath = settings.experimental.useFlatConfig ? 'eslint/use-at-your-own-risk' : 'eslint';
+			const eslintPath = ifUseFlatConfig? 'eslint/use-at-your-own-risk' : 'eslint';
 			if (nodePath !== undefined) {
 				promise = Files.resolve(eslintPath, nodePath, nodePath, trace).then<string, string>(undefined, () => {
 					return Files.resolve(eslintPath, settings.resolvedGlobalPackageManagerPath, moduleResolveWorkingDirectory, trace);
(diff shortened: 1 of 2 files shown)


server/src/eslint.ts
```diff
@@ -866,10 +866,12 @@ export namespace ESLint {
 				moduleResolveWorkingDirectory = settings.workingDirectory.directory;
 			}
 
+			const ifUseFlatConfig = settings.experimental.useFlatConfig && fs.existsSync('eslint.config.js')
```
koshic, line 869: it's a wrong behaviour - config file name is not restricted to 'eslint.config.js', user can provide anything via https://eslint.org/docs/latest/use/command-line-interface#-c---config.

koshic (changes requested):

kecrily:

server/src/eslint.ts
```diff
@@ -866,10 +866,16 @@ export namespace ESLint {
 				moduleResolveWorkingDirectory = settings.workingDirectory.directory;
 			}
 
+			const ifUseFlatConfig = settings.experimental.useFlatConfig
+				&& (
+					(settings.options?.overrideConfig && fs.existsSync(settings.options?.overrideConfig))
```
koshic, line 871: 1. I propose to change logic a bit and introduce 'flatConfigName' local variable - to avoid code duplication. like that:
```js
const flatConfigName = settings.options?.overrideConfig ?? 'eslint.config.js';
const ifUseFlatConfig = settings.experimental.useFlatConfig && fs.existsSync(flatConfigName);
```
2. Why 'settings.options?.' (expect undefined / null) but 'settings.experimental.' (expect that settings.experimental always defined)? May be we have some checks before, I just highlight the difference.

koshic (changes requested):

server/src/eslint.ts
kecrily: > Why 'settings.options?.' (expect undefined / null) but 'settings.experimental.' (expect that settings.experimental always defined)?

The default values are defined here. `experimental` has `useFlatConfig` as `false` by default. `options` has `{}` by default.

https://github.com/kecrily/vscode-eslint/blob/29ccab7bd909827b98c350933f21e6925c74697d/client/src/client.ts#L663-L672

But I found that the type definition of `options` (`ESLintOptions | undefined`) should be wrong. Based on the default value above, it can't be undefined anyway.

https://github.com/kecrily/vscode-eslint/blob/29ccab7bd909827b98c350933f21e6925c74697d/%24shared/settings.ts#L166

koshic (approved):

CvX:
Is there anything else that needs to be done before this PR can be merged? 😃🤞

koshic (approved):

kecrily:
Is it supposed to be automatic? Or should we mark it manually?

dbaeumer:
The PR got obsolute. ESLint added API that makes the detection of flat config a lot easier. To make use of this you need to have ESLint >= 8.57.0 and ESLint extension >= 3.0.5 pre-release.

PR #680, opened 2019-05-26
Added hideAutoFixedIssues option #267

kasvtv:
This implements #267, effectively allowing ESLint to be used as an unintrusive, silent, highly configurable, automatic formatter!

I took the liberty to slightly refactor the issue aggregation loop at [eslintServer.ts#L889](https://github.com/kasvtv/vscode-eslint/blob/b5bdd909697451ca81e0c5cd74d670b9cf7e2429/server/src/eslintServer.ts#L889) by extracting some self-containable logic (checking if issue needs autofixing) into a function [eslintServer.ts#L825](https://github.com/kasvtv/vscode-eslint/blob/b5bdd909697451ca81e0c5cd74d670b9cf7e2429/server/src/eslintServer.ts#L825), to make it clearer what's going on and more maintainable.

Remarks:
- This is my first PR to this repo, I couldn't find any tests anywhere, so I only tested manually, is this a mistake on my part?
- I came across a little confusing piece of logic. Fix actions were being recorded even when `problem.fix`, `problem.ruleId` or `cli.getRules` were unavailable. I might be mistaken, but that seems incorrect, isn't it? I changed it to not report on these conditions [eslintServer.ts#L826](https://github.com/kasvtv/vscode-eslint/blob/b5bdd909697451ca81e0c5cd74d670b9cf7e2429/server/src/eslintServer.ts#L826), otherwise it would've lead to very non-idiomatic code. Let me know if I have to revert this.

Thanks! Hope this makes it into a release. Would be great if I can reduce my tooling and unify all my style settings in one single .eslintrc!

Issue #267, opened 2017-07-19
Feature request: hide warnings/errors if they will be auto-fixed


ryaninvents:
Proposal: Add a new config option `eslint.hideAutoFixedIssues`.

If `eslint.hideAutoFixedIssues` and `eslint.autoFixOnSave` are both `true`, then do not show the error/warning in the text editor if ESLint suggests a fix. This is helpful when scanning a file, as it makes real errors (such as a typo in a variable name) much more visible.

I've got a bit of time coming up and would love to implement this.

dbaeumer:
Nice suggestion. A PR is welcome here. The file to look at is: https://github.com/Microsoft/vscode-eslint/blob/dbaeumer/multiroot/eslint-server\src\server.ts#L899

r24y:
Unfortunately it looks like if the error is reported, it _must_ be displayed, and if the error is not reported, it will not be auto-fixed. Bit of a Catch-22.

However, I was able to successfully downgrade issues with available fixes to `DiagnosticSeverity.Hint`, which provides a nice user experience. If the cursor is on the line, it displays as an unobtrusive lightbulb in the gutter, and if the cursor is on a different line there's no visual indication of the issue.

Do you think `hideAutoFixedIssues` is still an accurate enough phrase to describe this behavior? It's not really totally "hidden" per se, but all the other wordings I came up with sounded engineer-y and bizarre (`suppressAutoFixedIssues` was the closest I could come up with).

r24y:
I've made the PR, but it's open to maintainers, so you can modify the name of the option if need be.

dbaeumer:
@r24y thanks for looking into this. I think it should still be possible to hide the errors all together without downgrading them to a hint. The code flow should be something like this:

- recordCodeAction should return whether there was a fix or not
```ts
					if (problem) {
						let diagnostic = makeDiagnostic(problem);
						let hasFix = false;
						if (supportedAutoFixLanguages.has(document.languageId)) {
							hasFix = recordCodeAction(document, diagnostic, problem);
						}
						if (!(hasFix && hideAutofixableDiagnostics)) {
							diagnostics.push(diagnostic);
						}
					}

```

Have you tried this?

r24y:
@dbaeumer I haven't tried that. I might have time today to work on it

ajerez:
Any updates on this?

dbaeumer:
No, not from my side. @ryaninvents any update.

kasvtv:
[...]

PR #1705, opened 2023-08-27
Allow eslint.rules.customizations to target all fixable rules

### LINKED ISSUE #1634: Allow eslint.rules.customizations to target all fixable rules
#267 and #680 were closed through the implementation of the eslint.rules.customizations configuration option. However, it does not fully solve the use case of wanting to hide in-editor annotations for rules that will be autofixed - it would require manually adding overrides for specific rules, which is both time consuming/verbose and a moving target. It would be great if you could do something like: `[{"rule": "$fixable", "severity": "off"}]` or something similar

Dimava:
Closes #1634
Related: #267 , #680

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,14 +287,24 @@ This extension contributes the following variables to the [settings](https://cod
 
 - `eslint.rules.customizations` (@since 2.1.20) - force rules to report a different severity within VS Code compared to the project's true ESLint configuration. Contains two properties:
   - `"rule`": Select on rules with names that match, factoring in asterisks as wildcards: `{ "rule": "no-*", "severity": "warn" }`
-    - Prefix the name with a `"!"` to target all rules that _don't_ match the name: `{ "rule": "!no-*", "severity": "info" }`
+    - Prefix the name with a `"!"` to target all rules that _don't_ match the name: `{ "rule": "!no-*", "severity": "error" }`
+    - Suffix the name with `"@hasFix"` to target the errors which that have autofix available `{ "rule": "no-*@hasFix", "severity": "info" }`
+    - Suffix the name with `"!@hasFix"` to target the errors which that _don't_ have autofix available `{ "rule": "no-*@!hasFix", "severity": "error" }`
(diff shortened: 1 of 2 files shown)


Dimava:

README.md
```diff
@@ -287,14 +287,24 @@ This extension contributes the following variables to the [settings](https://cod
 
 - `eslint.rules.customizations` (@since 2.1.20) - force rules to report a different severity within VS Code compared to the project's true ESLint configuration. Contains two properties:
   - `"rule`": Select on rules with names that match, factoring in asterisks as wildcards: `{ "rule": "no-*", "severity": "warn" }`
-    - Prefix the name with a `"!"` to target all rules that _don't_ match the name: `{ "rule": "!no-*", "severity": "info" }`
+    - Prefix the name with a `"!"` to target all rules that _don't_ match the name: `{ "rule": "!no-*", "severity": "error" }`
+    - Suffix the name with `"@hasFix"` to target the errors which that have autofix available `{ "rule": "no-*@hasFix", "severity": "info" }`
+    - Suffix the name with `"!@hasFix"` to target the errors which that _don't_ have autofix available `{ "rule": "no-*@!hasFix", "severity": "error" }`
```
Dimava, line 292: Missed a !@hasFix type here

JoshuaKGoldberg, line 292: Instead of a suffix on the existing `rule` property, how about a new property like `"fixable"`? The suffix is confusing me and won't scale well if more features like it are asked for, such as filtering on whether it [provides _suggestions_](https://eslint.org/docs/latest/extend/custom-rules#providing-suggestions) or what it's [`meta.type`](https://eslint.org/docs/latest/extend/custom-rules#rule-structure) is.

Proposal:

```json
    { "rule": "*", "severity": "warn", fixable: true },
  ]
```

dbaeumer: @JoshuaKGoldberg I actually like this a lot. I am not really happy with the `!@hasFix` suffix either.

Dimava:
Implemented in #1841

PR #1770, opened 2024-01-22
Release 2.4.4.

dbaeumer:


benibenj (approved):

dbaeumer:
Since this is an extension we don't publish the extension to GitHub release page. It is only published in the market place. The reason for the GitHub releases are insider releases of the extension before VS Code had support for insider releases in the market place.

PR #1804, opened 2024-03-11
Hide all non-error messages when quiet mode is enabled

### LINKED ISSUE #1803: `eslint.quiet` should hide info level eslint rules as well
@ilias-t was very kind to add `eslint.quiet` option in https://github.com/microsoft/vscode-eslint/issues/637, which hides eslint warnings inside vscode. However, this does not hide eslint info level messages (which can be generated by using `eslint.rules.customizations.severity.downgrade` to downgrade warnings to info level).

Woodz:
Change logic to hide all non-error messages instead of only hiding warn level messages (which missed info level messages)

Resolve #1803

### DIFF
--- a/server/src/eslint.ts
+++ b/server/src/eslint.ts
@@ -1188,7 +1188,7 @@ export namespace ESLint {
 					docReport.messages.forEach((problem) => {
 						if (problem) {
 							const [diagnostic, override] = Diagnostics.create(settings, problem, document);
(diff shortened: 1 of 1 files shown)


dbaeumer:
This issue has been closed because it has not had recent activity. See also our [issue reporting](https://aka.ms/vscodeissuereporting) guidelines.

Happy Coding!

PR #1830, opened 2024-04-12
fix: add json, jsonc to `eslint.probe`'s default options

### LINKED ISSUE #1829: Update default options for `eslint.probe` to include json, jsonc
As we already support #1786

hyoban:
fix #1829

### DIFF
--- a/package.json
+++ b/package.json
@@ -320,7 +320,9 @@
 						"html",
 						"mdx",
 						"vue",
-						"markdown"
+						"markdown",
+						"json",
+						"jsonc"
 					],
 					"description": "An array of language ids for which the extension should probe if support is installed."
 				},


dbaeumer (approved):

jrieken (approved):

davidkarlsson:
These new defaults doesn't seem to be documented for the `eslint.probe` setting in README.md. 

The change caused a bit of confusion for me because vscode-eslint started showing a "No ESLint configuration found" error in 3.0.7 and above just because I happened to have a package.json file open from a project that shouldn't even have a eslint config because it's the root project of a yarn workspace.

dbaeumer:
Good catch. I updated the readme.

PR #1837, opened 2024-04-23
Performance improvements

DonJayamanne:
https://github.com/microsoft/vscode/issues/211142

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -9,7 +9,9 @@ import * as path from 'path';
 import {
 	workspace as Workspace, window as Window, languages as Languages, Uri, TextDocument, CodeActionContext, Diagnostic,
 	Command, CodeAction, MessageItem, ConfigurationTarget, env as Env, CodeActionKind, WorkspaceConfiguration, NotebookCell, commands,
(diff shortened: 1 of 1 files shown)


DonJayamanne:
Though this makes things fast, I do not believe this is the root cause,
This code path is getting hit when we execute cells in a notebook,
At this points the text documsnts of the cells do not change, hence I do not think eslint should fire.

Also shouldn't the condition here be changed from 
[if (Languages.match(packageJsonFilter, document) || Languages.match(configFileFilter, document) || validator.check(document) !== Validate.off) {](https://github.com/microsoft/vscode-eslint/blob/6f9512427253c2f50bd9c4ac171611452a779fdd/client/src/client.ts#L617)

to 

```typescript
if ((Languages.match(packageJsonFilter, document) || Languages.match(configFileFilter, document)) && validator.check(document) !== Validate.off) {
```

Perhaps I've miunserstood the code or just don't understand it, but if the cell language is Python I would have expected this path to bail early on, instead of even going into the `validator.check`.

dbaeumer:
Closing. Performance improvements have been implemented in the LSP libs.

Issue #633, opened 2019-02-25
Lint whole folder ignore workingDirectories setting


alexd16:
Hey

Not sure if this was already reported but could not find any similar issue.

I noticed that if I ran the 'lint whole folder' task it ignores the workingDirectories setting.
Is this intended behaviour?

If so is it something that would make sense to change? if so I could open a PR for it, would just need some pointer on how to approach it

Thanks

dbaeumer:
Honoring these setting would make perfect sense. The code should basically make a cd into these working dirs before executing the lint task. So instead of contributing one there would be n, one per working dir setting.

The corresponding code is here: https://github.com/Microsoft/vscode-eslint/blob/master/client/src/tasks.ts#L44

anmartins:
Hey,

I have a solution for this issue i'll open a PR.

Tiberriver256:
Was there a problem with @anmartins PR that prevented it from being merged? Looked like a pretty decent fix and would be nice to get this bug closed.

dbaeumer:
Actually I can't remember :-). Looking at it it actually looks good. If someone opens a new PR I will definitely have a look.

PR #1861, opened 2024-06-12
Added support for workingDirectories to the 'lint whole folder' task

Tiberriver256:
Fixes: #633

### DIFF
--- a/client/src/tasks.ts
+++ b/client/src/tasks.ts
@@ -40,8 +40,6 @@ class FolderTaskProvider {
 			return undefined;
 		}
 		try {
-			const command = await findEslint(rootPath);
-
 			const kind: EslintTaskDefinition = {
 				type: 'eslint'
 			};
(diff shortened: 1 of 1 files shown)


dbaeumer:
Nice work. Wasn't aware the `pushd` & `popd` is available in all command shells.

dbaeumer:
But what I found out now is that `&&` is not available under PowerShell so the task will fail if PowerShell is used as a command processor.

dbaeumer:
And `&&` usually has the semantic that the second command will not run if the first fails. IMO this will stop the shell execution on the first working directory that has an eslint error since eslint returns a exit code different than 0 in that case.

dbaeumer:
@Tiberriver256 any thoughts on the above comments.

Tiberriver256:
> But what I found out now is that `&&` is not available under PowerShell so the task will fail if PowerShell is used as a command processor.

Yes :( sadly older versions of PowerShell do not have this capability. Newer versions of PowerShell core do have support for it. We could look into using `;` instead but I'm not sure if that's supported in cmd.exe. It should work in Bash & PowerShell and would give us the behavior of running all the commands regardless of whether or not one failed.



> And `&&` usually has the semantic that the second command will not run if the first fails. IMO this will stop the shell execution on the first working directory that has an eslint error since eslint returns a exit code different than 0 in that case.

This is correct. Was a royal pain when I was testing it out locally to make sure I had run npm install in all my folders before it would run right.



> @Tiberriver256 any thoughts on the above comments.

Overall... feels flakey... I kinda want instead of one mega task, one task per folder. Instead of seeing this:

> - eslint whole folder

I would see:

> - eslint whole ./project1 folder
> - eslint whole ./project2 folder
> - eslint whole ./project3 folder

What do you think?

dbaeumer:
I will talk to the task developer in VS Code to see what we can do here.

dbaeumer:
Here is an idea you could experiment with:

- have a task per working directory
- contribute a command to execute all tasks. The implementation of the command could use the `workbench.action.tasks.runTask` API command documented here: https://code.visualstudio.com/api/references/commands

Let me know how it goes

Tiberriver256:
> Here is an idea you could experiment with:
> 
> * have a task per working directory
> * contribute a command to execute all tasks. The implementation of the command could use the `workbench.action.tasks.runTask` API command documented here: https://code.visualstudio.com/api/references/commands
> 
> Let me know how it goes

Nice! I'll give it a go.

PR #1918, opened 2024-08-12
Add mjs and cjs for config files.

dbaeumer:


bpasero (approved):

polyzen:
Seems https://github.com/microsoft/vscode-eslint/blob/ce243636ef106964ebee50626f73eac5a8c21bdb/client/src/client.ts#L427 also needs to be adjusted.

PR #1965, opened 2025-01-02
Probing support for Civet

edemaine:
This PR adds support for automatic eslint probing for the [Civet language](https://civet.dev) according to the existence of the plugin [eslint-plugin-civet](https://www.npmjs.com/package/eslint-plugin-civet). This avoids having to force linting via [`eslint.validate`](https://github.com/DanielXMoore/Civet/tree/main/integration/eslint#vs-code-configuration) (which also works but is more tedious for the user).

Here's an [example repo directory](https://github.com/DanielXMoore/Civet/tree/main/integration/eslint/example) to test on. Here's a screenshot of it working with this PR (lots of intentional errors):

image

Based on #1795 (analogous PR for Astro)


dbaeumer:
Can you please update the readme as well to outline that we have that support now.

dbaeumer (changes requested):
See previous comment

edemaine:
Sorry for the delay. I added `"civet"` to the default for `eslint.probe` in the README, which seems to be the only place these are mentioned. If you'd like me to add it to the two changelogs as well, I can do that — I assume for an odd-numbered prerelease version?

dbaeumer (approved):

alexdima (approved):

Issue #1917, opened 2024-08-09
Support TS config files


polyzen:
ESLint 9.9.0 was just released with support for TS config files:
https://github.com/eslint/eslint/pull/18134

polyzen:
Not sure if it's as simple as adding eslint.config.ts, eslint.config.mts, and eslint.config.cts to https://github.com/microsoft/vscode-eslint/blob/release/3.0.10/server/src/eslint.ts#L800

polyzen:
Need also to at least pass `--flag unstable_ts_config`, but not having success using a TS config with just ESLint myself yet.
https://eslint.org/docs/latest/use/configure/configuration-files#typescript-configuration-files

hyoban:
You can add the following settings:

```json
{
  "eslint.options": {
    "flags": ["unstable_ts_config"]
  }
}
```

> eslint.options: options to configure how ESLint is started using either the [ESLint class API](http://eslint.org/docs/developer-guide/nodejs-api#eslint-class) or the [CLIEngine API](http://eslint.org/docs/developer-guide/nodejs-api#cliengine).
> 
> options.flags (string[])
> Default is []. The feature flags to enable for this instance.

- https://eslint.org/docs/latest/integrate/nodejs-api#parameters
- https://github.com/Microsoft/vscode-eslint#settings-options

dbaeumer:
I added mjs and cjs to the watcher pattern. With the ts I would like to wait to see where the feature in eslint is heading to.

nzakas:
@dbaeumer what questions can I answer about this feature for you?

(In general, feel free to reach out via GitHub discussion or Discord if you're ever blocked waiting for info about what ESLint is doing.)

dbaeumer:
@nzakas my biggest question is currently why this is in ESLint since NodeJS itself added experimental support for this: https://nodejs.org/api/typescript.html#type-stripping

nzakas:
@dbaeumer because ESLint needs to run on multiple versions of Node.js, some of which don't have experimental support, as well as Bun and Deno. Once TypeScript support is built into Node.js as stable, and we are able to move off of support for older versions, we can remove the `jiti` dependency.

gorstak17:
Hello, eslint VS extensions keeps looking for .eslintrc even though we migrated to eslint v9 and now using flat config base eslint.config.mjs

Any help? @polyzen @nzakas @dbaeumer

dbaeumer:
[...]

PR #1968, opened 2025-01-09
Add Typescript configuration files detection

mustevenplay:


dbaeumer:
@mustevenplay thanks. It is my understand from the past that to be able to process TypeScript configuration files there needs to be used an experimental flag (see https://github.com/microsoft/vscode-eslint/issues/1917). Are you planning to add support for this as well?

davidenke:
> @mustevenplay thanks. It is my understand from the past that to be able to process TypeScript configuration files there needs to be used an experimental flag (see #1917). Are you planning to add support for this as well?

I just encountered an error with eslint 9.18.0, that the flag is no longer needed:

```bash
Oops! Something went wrong! :(

ESLint: 9.18.0

Error: The flag 'unstable_ts_config' is inactive: This flag is no longer required to enable TypeScript configuration files.
```

I came from 9.10.0, back then it _has been necessary_.

mustevenplay:
Will any change to this PR be necessary then? I understand that a flag was required in previous versions in order to be able to use these Typescript configuration files, but is the intention to keep it backwards compatible? Or should we simply move on and assume the usage of the newer versions? @dbaeumer @davidenke

davidenke:
AFAIK you still need [jiti](https://github.com/unjs/jiti) to be installed. Isn't it needed for this change as well?

dbaeumer:
@davidenke thanks for pointing this out. I wasn't aware of that.

I think that `jiti` needs to be specified in the `package.json` file and is nothing we need to handle in the extension. Otherwise validation wouldn't work in the terminal.

I think there is no need to make this backwards compatible. It would be fair IMO to ask users to use latest if they want TS config file support.

ENK0DED (approved):

dbaeumer (approved):

jrieken (approved):

aeschli (approved):

Issue #1156, opened 2021-01-08
Have to reload window often


rolandoldengarm:
We've recently switched from TSLint to ESLint in a large mono repo. After a while, about every 10-30 minutes, ESLint stops working and I have to reload the VSCode window. It will then no longer show new issues, and when I click CTRL + . on an issue it found before the crash, it shows "No code actions available".

When I check the Output console, I see these errors all the time:
```
(node:14432) UnhandledPromiseRejectionWarning: Error: Unhandled method eslint/confirmESLintExecution
    at c:\Users\r_old\.vscode\extensions\dbaeumer.vscode-eslint-2.1.14\server\out\eslintServer.js:1:53829
    at c:\Users\r_old\.vscode\extensions\dbaeumer.vscode-eslint-2.1.14\server\out\eslintServer.js:1:54123
    at Immediate.<anonymous> (c:\Users\r_old\.vscode\extensions\dbaeumer.vscode-eslint-2.1.14\server\out\eslintServer.js:1:54488)
    at processImmediate (internal/timers.js:439:21)
(node:14432) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:14432) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
Once it has crashed, and I try to do CTRL + . for the code actions, I see this in the output console:
```
[Error - 1:08:53 PM] Request textDocument/codeAction failed.
  Message: Unhandled method eslint/confirmESLintExecution
  Code: -32601 
  ```

It just crashed again, at the time it happened this was in the logs:
```
<--- Last few GCs --->

[26272:0000014600000000]   597028 ms: Scavenge 3793.3 (3815.5) -> 3788.0 (3817.0) MB, 5.3 / 0.0 ms  (average mu = 0.936, current mu = 0.863) allocation failure 
[26272:0000014600000000]   597060 ms: Scavenge 3794.6 (3817.0) -> 3789.4 (3818.2) MB, 5.2 / 0.0 ms  (average mu = 0.936, current mu = 0.863) allocation failure 
[...]

PR #1973, opened 2025-01-19
clarify where to set `eslint.runtime` and `eslint.execArgv` options

ShahinSorkh:
Coming from [this comment](https://github.com/microsoft/vscode-eslint/issues/1156#issuecomment-756603383), I spent a long time to figure out why these options are not available in the user settings json file. So, I think it's worth updating the docs.


dbaeumer:
Making this so WSL specific is IMO not the right approach (same statement for example also applies to the remotes like SSH and Dev Containers). So I think we should talk more general about a remote setup or alternatively list for the settings their scope (which for the two you mention is machine-overridable). See https://github.com/microsoft/vscode-eslint/blob/main/package.json#L359

dbaeumer (changes requested):
See comment before.

ShahinSorkh:
I think it's worth going with a small hint here and then listing the scope of the settings in another PR.

dbaeumer (changes requested):
A slightly more readable version of the text.

If you are using remote connections (e.g., WSL, Remote SSH, DevContainer, etc.) and don’t want to modify workspace-wide settings, you need to run the "Preferences: Open Remote Settings (JSON)" command via the Command Palette. Set the desired options there, and then reload the editor to ensure the linter server applies the changes.

sandy081 (approved):

PR #1977, opened 2025-02-05
Let the server determine cwd on its own

### LINKED ISSUE #1727: Prefer to use node from environment
https://github.com/microsoft/vscode-eslint/issues/1596#issuecomment-1792157160 mentioned:

> I would only use a node version different the one the come with VS Code if you need to (e.g. using a eslint plugin that requires some special node features)

@dbaeumer Why is the recommendation to use the one that comes with VS Code?

Most other linters and VS Code extensions I am aware of defaults to using the one "from environment".

Examples:
- https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff
`importStrategy` default is `fromEnvironment`.
- The Python extension by default loads the venv from environment.
- https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode does as well.
- etc. etc.

It makes more sense to me at least. Each project, old or new, comes with a specific Node version. Imagine you work in multiple open source repositories. Most have a `.nvmr` or a `.node-version` file. If I jump around between these, the Node version and thus implicitly the ESLint features changes depending on this.

To me the best solution would be to do like Ruff, where you fall back to the built-in one if you cannot find one from the environment:

<html>
<body>

importStrategy | fromEnvironment | Strategy for loading the ruff executable. fromEnvironment picks up Ruff from the environment, falling back to the bundled version if needed. useBundled uses the version bundled with the extension.
-- | -- | --

</body>
</html>

Could this be considered?

wagenet:
If `cwd` is not specified in the `ServerOptions` then the Language Server will use its own logic to determine the correct working directory, including the root of the workspace. This then allows `eslint.runtime` to be set to `"node"` which will be run in the root of the workspace. For projects using tools like Volta, this leads to it using the Volta-specified node version as expected.

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -399,8 +399,8 @@ export namespace ESLintClient {
 			}
 			const debugArgv = ['--nolazy', '--inspect=6011'];
 			const result: ServerOptions = {
-				run: { module: serverModule, transport: TransportKind.ipc, runtime, options: { execArgv, cwd: process.cwd(), env } },
-				debug: { module: serverModule, transport: TransportKind.ipc, runtime, options: { execArgv: execArgv !== undefined ? execArgv.concat(debugArgv) : debugArgv, cwd: process.cwd(), env } }
+				run: { module: serverModule, transport: TransportKind.ipc, runtime, options: { execArgv, env } },
+				debug: { module: serverModule, transport: TransportKind.ipc, runtime, options: { execArgv: execArgv !== undefined ? execArgv.concat(debugArgv) : debugArgv, env } }
 			};
 			return result;
 		}


dbaeumer:
This is for security reasons not so easy. I need to think about the consequences of this and if trusted workspace allow this now. Using the extension host's CWD ensures that no unwanted binary code is loaded.

dbaeumer:
What we can always do is to have a opt-in setting for this.

wagenet:
@dbaeumer opt-in is ok for me. My goal is very much for the process to run in the context of my workspace.

wagenet:
@dbaeumer do you have any guidance on how to implement this. I'd still love to see this happen but my understanding of the codebase is pretty limited.

thernstig:
@wagenet this does not solve https://github.com/microsoft/vscode-eslint/issues/1727. Can you unlink it?

shah20033:
[index.html](https://github.com/user-attachments/files/30539786/index.html)

dbaeumer:
@wagenet are you still working on the PR?

wagenet:
@dbaeumer apologies for the silence here. I am not actively working on this.

Issue #1995, opened 2025-03-20
Bug: fix Unicode when saving


Mori-Yang:
In this [issue](https://github.com/antfu/eslint-plugin-command/issues/26), use `eslint . --fix`can run successfully, but using the eslint extension to fix it when saving will cause incorrect Unicode encoding.

I think this is a problem caused by the extension and hope to be answered and helped.

dbaeumer:
@Mori-Yang  Can you please provide me with a GitHub repository I can clone with a minimal setup that demos what you are seeing.

Mori-Yang:
Hi @dbaeumer , I created a repository:

https://github.com/[Mori-Yang/eslint-extension-test](https://github.com/Mori-Yang/eslint-extension-test#)#

you can try saving in `index.js`, the autofix will cause wrong Unicode. Then try using `eslint . --fix`, it works fine.

Mori-Yang:
Hi @dbaeumer .
Sorry to bother you, any progress here? I want to know if this problem is caused by the extension.

dbaeumer:
Thanks for the test case. I can reproduce this and I think it is a bug in the extension. Very likely a problem with the diffing algorithm.

Mori-Yang:
@dbaeumer  If you can tell me some modifications and code locations that maybe I can try, or if you think you can easily fix it, all the better!

dbaeumer:
The relevant code is here: https://github.com/microsoft/vscode-eslint/blob/c0e753713ea9935667e849d91e549adbff213e7e/server/src/eslintServer.ts#L748

I guess it has something to do with diffing here: https://github.com/microsoft/vscode-eslint/blob/c0e753713ea9935667e849d91e549adbff213e7e/server/src/eslintServer.ts#L790 which doesn't handle the unicode characters correctly when computing the edits.

dbaeumer:
Best you can do is to beug it and see where exactly the problem happens. Ping me if you have more questions.

Mori-Yang:
Thank you, I'll try.

Mori-Yang:
> Best you can do is to beug it and see where exactly the problem happens. Ping me if you have more questions.

Hi, @dbaeumer . I debugged the code you mentioned and found that there is no problem with the stringDiff function. Now I understand that the cause of the problem may be:
[...]

PR #2007, opened 2025-04-18
WIP : Unicode diff error

Mori-Yang:
issue #1995 

Sorry to bother you, just take a look when you have time.

Now I modified the diff and [positionAt](https://github.com/microsoft/vscode-languageserver-node/pull/1625) codes and can get the correct `result`：
image

but it still shows the wrong result，I may need further help from you.


dbaeumer:
@Mori-Yang thanks. Could you explain what you think the problem is  and what your fix does?

PR #2031, opened 2025-06-28
chore: fix typo in comment-out

noritaka1166:
fix typo in comment-out


dbaeumer (changes requested):
@noritaka1166 thanks. Actually the right wording fix should be `This can only happen if size === 1 which is handled`

noritaka1166:
> @noritaka1166 thanks. Actually the right wording fix should be `This can only happen if size === 1 which is handled`

@dbaeumer 
Done!

DonJayamanne (approved):

PR #2032, opened 2025-06-28
refactor: replace substr with slice & use startsWith and endsWith

noritaka1166:
I did the following refactoring.
- replace deprecated substr with slice
- use startsWith and endsWith for better readability
- fix lint warning

### DIFF
--- a/server/src/eslintServer.ts
+++ b/server/src/eslintServer.ts
@@ -136,7 +136,7 @@ function inferFilePath(documentOrUri: string | TextDocument | URI | undefined): 
 					const extension = LanguageDefaults.getExtension(textDocument.languageId);
 					if (extension !== undefined) {
 						const extname = path.extname(filePath);
-						if (extname.length === 0 && filePath[0] === '.') {
+						if (extname.length === 0 && filePath.startsWith('.')) {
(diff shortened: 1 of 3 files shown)


server/src/eslintServer.ts
```diff
@@ -136,7 +136,7 @@ function inferFilePath(documentOrUri: string | TextDocument | URI | undefined):
 					const extension = LanguageDefaults.getExtension(textDocument.languageId);
 					if (extension !== undefined) {
 						const extname = path.extname(filePath);
-						if (extname.length === 0 && filePath[0] === '.') {
+						if (extname.length === 0 && filePath.startsWith('.')) {
```
dbaeumer, line 139: Why is this better than the array access?

dbaeumer (changes requested):

server/src/eslintServer.ts
noritaka1166: @dbaeumer 
SonarQube, which I often use, says [this](https://next.sonarqube.com/sonarqube/coding_rules?open=typescript%3AS6557&rule_key=typescript%3AS6557), but do you prefer array access?

スクリーンショット 2025-08-17 午後10 59 10

PR #2076, opened 2025-09-21
Detect root via lockfiles; do not treat package.json as root

kapral18:
## Summary
- Stop marking `package.json` as a project root by itself
- Treat lockfiles (`package-lock.json`, `yarn.lock`, `pnpm-lock.yaml`, `npm-shrinkwrap.json`, `bun.lockb`) as roots
- Preserve flat config priority via existing `eslint.config.*` entries

This ensures auto working directory detection stops at real project roots while allowing nested `package.json` without lockfiles to bubble up.


dbaeumer:
@kapral18 thanks for the PR.

Should the logic not be like this:

- if it has a packag.json and one of the lock files take it as the root
- if it has only a package.json traverse upwards. In not better match is found treat that one as the root.

kapral18:
> @kapral18 thanks for the PR.
> 
> Should the logic not be like this:
> 
> * if it has a packag.json and one of the lock files take it as the root
> * if it has only a package.json traverse upwards. In not better match is found treat that one as the root.

yes that + logic for flat config files also signaling root, would be a better root detection probably, especially considering that there can be eslintConfig blocks inside package.json (which are no longer supported in v9+ but still)

kapral18:

kapral18:
@dbaeumer I split the indicators into semantic groups so that the priority and connection is a little clearer to read and easier to work with. Priority is handled a little more explicitly I think now, but check out and let me know what you think? I didn't add test infrastructure to avoid exploding this PR's size since there is no testing infra for server available in the first place. But I tested locally so let me know if this can work

kapral18:
@dbaeumer actually I just realized that this very repository is breaking my assumptions about eslint.config.js usage. In my proposal I am looking at eslint.config.js as a primary boundary indicator superior to lock files, because flat config files like eslint.config.js are supposed to be used in project roots, but here you have lock roots and eslintrc.js root file in server and client respectively while putting an eslint.config.js on the .git root level. Can you kindly help me understand what is the grand design here?

dbaeumer:
@kapral18 the `eslintrc.json` files are obsolete and not needed anymore. I deleted them.

kapral18:
@dbaeumer I still see them in https://github.com/microsoft/vscode-eslint/blob/main/server/src/eslint.ts#L816 or what do you mean? Do you mean VSCode extension no longer supports .eslintrc.json files specifically?

dbaeumer:
@kapral18 sorry I miss your last comment. I misunderstood you. The repository itself had `eslintrc.json` file in it which were obsolete and I deleted them. The once in the source code handle the old non flat config files which usually don't indicate the root of a project precisely. That is why isRoot is false.

dbaeumer:
@kapral18 can you fill me in on the state of this PR?

PR #2081, opened 2025-09-28
support setting command in config for lintTask

### LINKED ISSUE #1827: [feat] `eslint.lintTask.command: string[]`
https://github.com/microsoft/vscode-eslint/blob/main/client/src/tasks.ts#L43-L56

https://github.com/microsoft/vscode-eslint/blob/main/client/src/node-utils.ts#L56-L65

Seems like there is no option to configure eslint lint task executable path.

For example, when using `yarn@4.1.1`, there is no `node_modules`.

Would be nice to have option to configure task command prefix array

```jsonc
"eslint.lintTask.command": ["corepack", "yarn", "eslint"],
```

frodi-karlsson:
Closes #1827 

Just tested this in a pnpm workspace using `pnpm eslint` as the command, but seems straight forward enough.


dbaeumer:
Otherwise a nice PR

Issue #1754, opened 2023-12-15
ESLint client: couldn't create connection to server.


chrispalmo:
# Issue Summary for vscode-eslint

## Error Description
While using the ESLint extension in VSCode, the following errors are encountered:

- ESLint server starts and stops immediately.
- Error message: "couldn't create connection to server."
- Issue with launching server using runtime node: "Error: spawn node ENOENT."

```
[Info  - 12:28:41] ESLint server is starting.
[Info  - 12:28:41] ESLint server stopped.
[Error - 12:28:41] ESLint client: couldn't create connection to server.
Launching server using runtime node failed. Error: spawn node ENOENT
[Error - 12:28:41] Starting the server failed.
Launching server using runtime node failed. Error: spawn node ENOENT
[Info  - 12:28:41] PATH environment variable is: /usr/bin:/bin:/usr/sbin:/sbin
```

## Diagnostic Information
- MacOS Sonoma 14.2 (23C64), Apple M1 Max
- Node.js version: v18.18.2
- Location of Node.js: `/Users/cp/.npm/bin/node`

## Steps Taken to Resolve
1. **Node.js Installation Verification:** Confirmed Node.js is installed and accessible via PATH.
2. **Updated PATH Variable:** Verified Node.js path is included in the system's PATH.
3. **VSCode Settings for ESLint:**
   - Tried various configurations for `eslint.nodePath` including empty, specific path and null.
4. **Reinstall ESLint Extension:** Uninstalled and reinstalled the ESLint extension without resolution.
5. **Check and Update Node Version:** Using `n`, updated and verified the Node version.
6. **ESLint Runtime Path Setting:** Attempted setting `eslint.runtime` to the path of the Node executable.
7. **User Settings Configuration:** Tested with different configurations in VSCode user settings.
8. **Permissions Check:** Verified permissions for Node binary and relevant directories.
9. **Direct Execution in VSCode Terminal:** Confirmed ability to run Node via VSCode terminal.
10. **Setting `eslint.runtime` to Null:** Tried explicitly setting `eslint.runtime` to null.
11. **Permissions and Configuration:**
  - Explored the possibility of permission issues with the Node binary and relevant directories.
  - Checked and confirmed proper read/write/execute permissions.
  - VSCode user configuration was modified and tested in various states, including removing all properties, without resolving the issue.

## Additional Context
- The Node.js installation was managed using `n` (version 8.2.0).
- Fresh installation of VSCode
[...]

PR #2087, opened 2025-10-16
Configure shell env timeout to avoid PATH problems on VS Code update

karlhorky:
On VS Code Update, errors appear:

- https://github.com/microsoft/vscode-eslint/issues/1754#issuecomment-3410174695

Instead, suggest users configure a longer shell environment resolution timeout of 30 seconds via  `"application.shellEnvironmentResolutionTimeout": 30`, as mentioned here:

- https://github.com/microsoft/vscode-eslint/issues/1754#issuecomment-3410201863

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,18 @@
+### Version 3.0.20 - pre-release
+
+- move to latest LSP libraries
+- `eslint.quite` blocks info messages as well.
+
+### Version 3.0.20 - release
+
+- [Bug fixes](https://github.com/microsoft/vscode-eslint/issues?q=is%3Aissue%20state%3Aclosed%20milestone%3A3.20.0)
+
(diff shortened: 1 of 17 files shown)


dbaeumer:
This sounds like a pre-requisite. I would just point to the setting saying something like if you see the message Unable to resolve your shell environment in a reasonable time.  use the setting to enlarge the timeout

karlhorky:
@dbaeumer thanks for the feedback.

What do you think about putting it in a "NOTE" bullet?

- [`3c335b9` (#2087)](https://github.com/microsoft/vscode-eslint/pull/2087/commits/3c335b9ac1e0bfa8273fab59c48d28633243e03a)

dbaeumer:
Works for me :-)

dbaeumer:
@microsoft-github-policy-service rerun

PR #2159, opened 2026-03-30
Support VSCode on the Web

### LINKED ISSUE #1366: Support for VS Code for Web
The 'ESLint' extension is not available in Visual Studio Code for the Web.

[https://vscode.dev](https://vscode.dev/)

Learn Why [https://code.visualstudio.com/docs/editor/vscode-web#_extensions](https://code.visualstudio.com/docs/editor/vscode-web#_extensions)

ChaseKnowlden:
Fixes #1366

### DIFF
--- a/browser-stubs/path.js
+++ b/browser-stubs/path.js
@@ -1,0 +1,95 @@
+/* Browser stub for the 'path' module used by ESLint's Linter class */
+'use strict';
+
+const sep = '/';
+const delimiter = ':';
+
+function normalize(p) {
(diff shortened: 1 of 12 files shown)


dbaeumer:
@ChaseKnowlden thanks for the PR. Since this is more or less a different implementation that we have in the desktop I thik we need to be very explicit about its feature set and which `eslint.*` settings will work in the desktop, in the web and in both. Otherwise it will be very confusing for users.

PR #2170, opened 2026-05-16
Log restart success message to output channel

### LINKED ISSUE #1792: Add restart message to Output console message to confirm restart
When using the output channel to diagnose problems, there is no visual indication when the problem gets fixed. 

As a simple example, here is a new project where I forgot to installed eslint:
image

What you can't tell is that I went and fixed the problem by installing eslint and the eslint extension is happy and running as expected. However, it looks like the problem is still active because there has been no change to the console.

Now imagine diagnosing a more difficult setup issue, and you keep getting the same error as you try various fixes. And then you actually fix the problem but you move past it because you can't tell it worked. Assuming you notice what's going on, you still have to stop and manually go try and see if linting is working, or you just restart VSCode to figure out correct status. Good times.

A single line on successful restart would be very helpful!

singhvishalkr:
When diagnosing ESLint setup problems, I noticed there's no visual indication in the Output console when a restart succeeds. After fixing an issue (like installing the missing eslint package) and running "ESLint: Restart ESLint Server", the console still shows the old error with no follow-up message confirming that things are now working.

I added a simple `client.info('ESLint server restarted.')` line after both restart paths succeed. Now users get a clear signal that the server came back up without having to manually check if linting is working.

Closes #1792

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -146,7 +146,9 @@ function realActivate(context: ExtensionContext): void {
 			if (client.state === State.StartFailed) {
 				await client.dispose();
 				[client, acknowledgePerformanceStatus] = ESLintClient.create(context, validator);
-				return client.start().catch((error) => {
+				return client.start().then(() => {
+					client.info('ESLint server restarted.');
+				}).catch((error) => {
 					client.error(`Starting the server failed.`, error, 'force');
 					const message = typeof error === 'string' ? error : typeof error.message === 'string' ? error.message : undefined;
 					if (message !== undefined && message.indexOf('ENOENT') !== -1) {
@@ -154,7 +156,9 @@ function realActivate(context: ExtensionContext): void {
 					}
 				});
 			} else {
-				return client.restart().catch((error) => client.error(`Restarting client failed`, error, 'force'));
+				return client.restart().then(() => {
+					client.info('ESLint server restarted.');
+				}).catch((error) => client.error(`Restarting client failed`, error, 'force'));
 			}
 		}),
 		Commands.registerCommand('eslint.revalidate', () => {


dbaeumer:
@singhvishalkr thanks for the PR. Can you provide (a screen shot or the text) how this now looks in the output channel.

singhvishalkr:
@dbaeumer After either restart path finishes successfully, the ESLint language client writes an info line to the **ESLint** output channel. The message body is exactly `ESLint server restarted.` (the string passed to `client.info` in `client/src/extension.ts`). Typical presentation is the channel's standard info formatting around that substring; older errors above stay in the scrollback so successful reload reads as one clear tail line rather than silence.

dbaeumer:
There is already output like this

```
2026-05-18 15:04:44.753 [info] ESLint server is starting.
2026-05-18 15:04:44.785 [info] ESLint server running in node v22.22.1
2026-05-18 15:04:44.787 [info] ESLint server is running.
2026-05-18 15:04:45.115 [info] ESLint library loaded from: xxx/node_modules/eslint/lib/api.js
```

Which prints that the ESLint server is running. I am wondering why we need a started message and where it fits into the log

singhvishalkr:
@dbaeumer The lines you quoted are the cold-start sequence. They do not appear when someone already has a failed ESLint server, runs `ESLint: Restart ESLint Server`, and only needs a deterministic tail line that the restart path finished. The restart command reuses the channel but does not repeat the full boot banner, so without an explicit `client.info` the last visible line can still be the stale failure. The added line is scoped to the two successful restart entry points only.

dbaeumer (approved):

dbaeumer:
You are right.

Yoyokrazy (approved):

PR #2172, opened 2026-05-18
Refactor experimental flat config handling

jpc-ae:
Prevent spurious `ESLint version 10.x.x only supports flat configs. Setting is ignored.` warnings by only adding the experimental setting if it exists in the file.

Closes Issue #2138

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -689,14 +689,20 @@ export namespace ESLintClient {
 					}
 				});
 				const useFlatConfig = config.get<boolean | null>('useFlatConfig', null);
+				const useExperimentalFlatConfig = config.get<boolean | null>('experimental.useFlatConfig', null);
 				const settings: ConfigurationSettings = {
 					validate: Validate.off,
 					packageManager: config.get<PackageManagers>('packageManager', 'npm'),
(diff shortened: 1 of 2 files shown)


dbaeumer:
@jpc-ae thanks for the PR. IMO a better fix would be to remove `settings.experimental?.useFlatConfig === false` from here https://github.com/microsoft/vscode-eslint/blob/dbaeumer/wrong-egret-yellow/server/src/eslint.ts#L998 sind the experimental setting was only useful in version before 8.57.x.

dbaeumer (changes requested):
See comment before

jpc-ae:
That was my first thought, but the way it was written made me think the author was trying to be explicit. I will try again.

jpc-ae:
According to [the type definition](https://github.com/jpc-ae/vscode-eslint/blob/26e8522035bd715f6c50007fc7e1ed804ad63368/%24shared/settings.ts#L172), experimental can be undefined. I've updated my patch to make it cleaner according to the existing coding style. I've also updated the check to be more explicit and useful for users.

jpc-ae:

dbaeumer (approved):

alexr00 (approved):

PR #2175, opened 2026-05-18
Ignore rule if file is dirty

### LINKED ISSUE #2171: Ignore rule if file is dirty
I'm using `eslint-plugin-prettier` for CI (e.g. pull requests) to check if new code is formatted according to how Prettier formats.

When editing code however, the plugin's accompanying `prettier/prettier` rule is annoying because the point of auto-formatters is that you can just write code without formatting yourself and the code will be auto-formatted on save or similar.

Luckily this extension has an option to turn off rules like this:
```json
	"eslint.rules.customizations": [{ "rule": "prettier/prettier", "severity": "off" }]
```

This works great, but what if I'm using something like codemods, codegen tools, or anything else that modifies or writes code without ESLint / Prettier integration? I'd need to remember to run `prettier . --write` or ESLint CLI - or find out the code isn't formatted when the code runs through CI.

Instead, if the `.vscode/settings.json` rule customization only applied to non-dirty files, issues of incorrect formatting (or other ESLint rules) would be visible much sooner but still won't be annoying while editing the file (when the file is dirty / unsaved).

I'm proposing a new option for `eslint.rules.customizations` entries like `unsavedOnly?: boolean`.

FunctionDJ:
Closes #2171.

Adds an `unsavedOnly` boolean property to `eslint.rules.customizations` entries. When `true`, the customization only applies while the file has unsaved changes; once saved, the rule reports at its original severity.

**Motivating use case:** `eslint-plugin-prettier` is useful in CI but noisy during active editing. With this option you can suppress `prettier/prettier` while typing while still seeing violations in committed/saved code:

```json
  { "rule": "prettier/prettier", "severity": "off", "unsavedOnly": true }
]
```

**Implementation notes:**

Dirty state is tracked server-side by recording the document version at open/save (`textDocument/didOpen`, `textDocument/didSave`) and comparing against the current version. This avoids a false-dirty state that would arise from using `TextDocuments.onDidChangeContent`, which fires for both `didOpen` and `didChange` notifications.

Because the pull diagnostics model doesn't automatically re-pull after a save when in-memory content is unchanged, `onDidSave` triggers `connection.languages.diagnostics.refresh()` so the updated dirty state is reflected immediately.

The `RuleSeverities` cache key is extended with the dirty flag (`ruleId:0/1`) so clean and dirty results are cached independently and invalidated correctly on settings change.

Human written AI Disclaimer: I used Copilot to help me with this implementation and the above PR text body, but I reviewed it, ran the mocha tests, used the Launch Config to test it, and iterated on the feature until it actually worked with Launch Config. I take responsibility for this code. I also looked into if there's a simpler way to determine if a document is dirty, but according to Copilot this would require a client-server roundtrip that's not necessary when the implementation in this PR also works server-side-only.

### DIFF
--- a/$shared/settings.ts
+++ b/$shared/settings.ts
@@ -121,6 +121,8 @@ export type RuleCustomization = {
 	severity: RuleSeverity;
 	/** Only apply to autofixable rules */
 	fixable?: boolean;
+	/** Only apply when the file has unsaved changes */
+	unsavedOnly?: boolean;
 };
 
 export type RunValues = 'onType' | 'onSave';
--- a/README.md
+++ b/README.md
@@ -432,6 +432,7 @@ This extension contributes the following variables to the [settings](https://cod
     - Prefix the name with a `"!"` to target all rules that _don't_ match the name: `{ "rule": "!no-*", "severity": "info" }`
   - `"severity"`: Sets a new severity for matched rule(s), `"downgrade"`s them to a lower severity, `"upgrade"`s them to a higher severity, or `"default"`s them to their original severity
   - `"fixable"`: Select only autofixable rules: `{ "rule": "no-*", "fixable": true, "severity": "info" }`
+  - `"unsavedOnly"`: When `true`, only applies the customization while the file has unsaved changes (is dirty): `{ "rule": "...", "severity": "off", "unsavedOnly": true }`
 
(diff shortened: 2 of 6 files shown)


FunctionDJ:

PR #2206, opened 2026-07-20
Show ESLint output after manual restart

### LINKED ISSUE #2205: Show output channel on manual restart
When changing config one often has to restart the extension: `"ESLint: Restart ESLint Server"`.

It would be very helpful if that would automatically open the "output" channel, so one can inspect logs - since a config change was made, one can thus ensure there are no bugs, and generally note things are working correctly.

This would annoy some users, so could be hidden by a setting which is disabled by default, e.g. `"eslint.showOutputOnRestart": false`.

Thanks!

pradeep-ramola:
**Status:** Draft pending microsoft/vscode-languageserver-node#1849. Once the shared `LanguageClient.restart()` behavior lands and is available through `vscode-languageclient`, this ESLint PR can be simplified to rely on the client library behavior instead of carrying an ESLint-specific restart implementation.

Fixes microsoft/vscode-eslint#2205

## Summary

Preserves the ESLint output channel when users manually restart the ESLint server, without adding a new setting.

If the ESLint output channel is already visible before `ESLint: Restart ESLint Server` runs, the extension shows it again after the restart succeeds. If the output channel is not visible, restart behavior stays unchanged.

## Changes

- Removes the proposed `eslint.showOutputOnRestart` setting and README entry.
- Detects whether the ESLint output channel is visible before restart using VS Code's visible text editors for output documents.
- Restores the output channel with `show(true)` after successful restart paths, preserving focus.

## Testing

- `npm run lint`
- `npm run lint --prefix client`

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,6 +371,7 @@ This extension contributes the following variables to the [settings](https://cod
 
 - `eslint.enable`: enable/disable ESLint for the workspace folder. Is enabled by default.
 - `eslint.debug`: enables ESLint's debug mode (same as --debug  command line option). Please see the ESLint output channel for the debug output. This options is very helpful to track down configuration and installation problems with ESLint since it provides verbose information about how ESLint is validating a file.
+- `eslint.showOutputOnRestart`: shows the ESLint output channel after manually restarting the ESLint server. Disabled by default.
 - `eslint.lintTask.enable`: whether the extension contributes a lint task to lint a whole workspace folder.
 - `eslint.lintTask.options`: Command line options applied when running the task for linting the whole workspace (https://eslint.org/docs/user-guide/command-line-interface).
   An example to point to a custom `.eslintrc.json` file and a custom `.eslintignore` is:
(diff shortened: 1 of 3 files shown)


dbaeumer:
I am not a fan of yet another setting. If the output channel is visible then it will be preserved after a restart.

pradeep-ramola:
Updated as requested. I removed the new `eslint.showOutputOnRestart` setting and its README/package contribution.

The restart command now checks whether the ESLint output channel is already visible before restarting. If it was visible, the extension restores it after a successful restart with `show(true)` so focus is preserved. If it was not visible, restart behavior remains unchanged.

Validated with:

- `npm run lint`
- `npm run lint --prefix client`

dbaeumer:
This would also be better implemented in the LSP client libraries so that every client can profit from that. The restart method is here: https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/node/main.ts#L196

Would you be willing to have a look.

pradeep-ramola:
Yes, I had a look and opened microsoft/vscode-languageserver-node#1849 to move the behavior into `LanguageClient.restart()`.

I also converted this PR to draft while the shared client-library change is reviewed. Once that lands and flows into `vscode-languageclient`, I’ll update this ESLint PR to rely on the library behavior instead of carrying an ESLint-specific implementation.

PR #2225, opened 2026-08-28
#2224 - use HTML comments for disable rule quick fixes in Markdown files

### LINKED ISSUE #2224: Disable rule quick fixes insert JS-style comments into Markdown files
Hello, thanks for maintaining this extension!
 
I found that the disable-rule quick fixes insert JS-style comments into Markdown files, where they are just literal text — the diagnostic is not suppressed. Markdown needs HTML comments: `<!-- eslint-disable-next-line {rule} -->`.

> I'd like to work on this — I'll follow up with a PR.

Image

Image

Image

## Repro

`eslint@9` + `@eslint/markdown@7`, any `.md` file with a violation:

```js
  // eslint.config.mjs
  import markdown from '@eslint/markdown';

  export default [{
        files: ['**/*.md'],
        plugins: { markdown },
        language: 'markdown/gfm',
        rules: { 'markdown/fenced-code-language': 'error' }
  }];
```

- Lightbulb → "Disable ... for this line" inserts `// eslint-disable-next-line ...` and the error remains. "Disable for the entire file" is affected too (`/* eslint-disable ... */`). I verified via CLI that the HTML-comment forms do suppress the diagnostics.

## Cause
- `server/src/languageDefaults.ts` has no `markdown` entry in `languageId2Config`, so the comment style falls back to `//` / `/* */`.

## Fix
- Add a `markdown` entry with `blockComment: ['<!--', '-->']` (no line comment).
- Fall back to block style when the language has no line comment, regardless of `eslint.codeAction.disableRuleComment.commentStyle`.

Related: #1074, #2039

dev-oil:
Fixes #2224

  Markdown has no `//` or `/* */` comment syntax, so the disable-rule quick fixes
  inserted literal text that didn't suppress anything.

  - Added a `markdown` entry to `languageId2Config` with `blockComment: ['<!--', '-->']` and no line
  comment.
  - The disable code actions now fall back to a block comment when the language has no line comment,
  regardless of the `commentStyle` setting.

  Note: adding the entry also makes `getExtension('markdown')` return `md` (previously `undefined`), so
  untitled markdown documents and notebook cells now get a virtual `.md` path like other languages.

  Happy to adjust if you prefer a different approach.


dev-oil:

PR #2227, opened 2026-08-31
Add option to ignore files outside the workspace

### LINKED ISSUE #1805: Do not lint outside current project workspace directory
Maybe related: https://github.com/microsoft/vscode-eslint/issues/196
https://www.reddit.com/r/learnjavascript/comments/xsiowg/i_need_some_help_configuring_my_eslint_correctly/
https://stackoverflow.com/questions/49422058/vscode-eslint-not-ignoring-directory

The problem is that extension keeps linting files outside current project's root directory.
For example if I have 
`D:\projects\A` (vscode cwd) [${workspaceFolder}](https://code.visualstudio.com/docs/editor/variables-reference#_predefined-variables)
`D:\projects\B` (eslint config defined) (and a file opened in `A` project)

QinXi-ai:
## Problem

Opening a local file outside the current workspace can make the extension load and run an unrelated ESLint configuration. The existing behavior is useful for single-file workflows, so changing it unconditionally would break compatibility.

## Design

- Add an opt-in `eslint.ignoreOutsideWorkspace` resource setting, defaulting to `false`.
- When enabled and at least one workspace folder is open, skip local `file:` documents that do not belong to any workspace folder.
- Keep untitled and virtual documents unchanged, and keep standalone-file validation working when no folder is open.
- Apply the boundary before explicit `eslint.validate` rules so the option consistently blocks external files.
- Re-evaluate synced documents when workspace folders are added or removed, in addition to configuration changes.
- Extract the validator behind a small injectable workspace interface so boundary behavior can be tested without a VS Code host. Preserve legacy `eslint.validate` entry validation, including rejecting non-boolean `autoFix` values.

## Tests

- `npm run compile`
- `npm run lint --prefix client`
- `npm run lint --prefix server`
- `npm test` (client: 25 passed; server: 5 passed)
- `git diff --check`
- Regression tests reject string, null, numeric, object, and array `autoFix` values; these seven cases failed before the guard fix and pass afterward. Valid legacy entries and subsequent valid language entries remain accepted.

Fixes #1805


QinXi-ai:

client/src/validator.ts
```diff
@@ -0,0 +1,92 @@
...
+
+function isValidateItem(item: unknown): item is ValidateItem {
+	return typeof item === 'object' && item !== null && 'language' in item && Is.string(item.language);
+}
+
+export interface ValidatorConfiguration {
+	get<T>(section: string, defaultValue: T): T;
+}
+
+export interface ValidatorWorkspace {
```
dbaeumer: The should be better

type ValidateWorkspace = Pick<typeof Workspace, 'workspaceFolders' | 'getConfiguration' | 'getWorkspaceFolder'>;

client/src/validator.ts
```diff
@@ -0,0 +1,92 @@
...
+		if (!config.get<boolean>('enable', true)) {
+			return Validate.off;
+		}
+
+		if (textDocument.uri.scheme === 'untitled' && config.get<boolean>('ignoreUntitled', false)) {
+			return Validate.off;
+		}
+
+		if (
+			textDocument.uri.scheme === 'file' &&
```
dbaeumer: Why do we limit this to `file`.

dbaeumer (changes requested):

QinXi-ai:
Addressed both review points in 70c63af:

- replaced the hand-written workspace interfaces with a Pick of the VS Code workspace API
- applied ignoreOutsideWorkspace to remote and virtual workspace documents while leaving untitled documents under ignoreUntitled
- added remote in-workspace and outside-workspace coverage and updated the setting documentation

Validation: client TypeScript compile passed, all 9 client tests passed, and client/server lint passed.

client/src/client.ts
```diff
@@ -190,7 +138,11 @@ export namespace ESLintClient {
 					provider?.send(textDocument).catch((error) => client.error(`Sending open notification failed.`, error));
 				}
 			}
-		}));
+		};
+		context.subscriptions.push(
+			Workspace.onDidChangeConfiguration(refreshSyncedDocuments),
+			Workspace.onDidChangeWorkspaceFolders(refreshSyncedDocuments)
```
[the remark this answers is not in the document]
dbaeumer: @Copilot Are you sure about this. Can you double check the implementation here: https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/features.ts#L345 and https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/textSynchronization.ts#L49

send on the provider calls the callback method which send everything through the middleware

client/src/validator.ts
```diff
@@ -0,0 +1,84 @@
...
+		if (!config.get<boolean>('enable', true)) {
+			return Validate.off;
+		}
+
+		if (textDocument.uri.scheme === 'untitled' && config.get<boolean>('ignoreUntitled', false)) {
+			return Validate.off;
+		}
+
+		if (
+			textDocument.uri.scheme !== 'untitled' &&
```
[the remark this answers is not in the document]
QinXi-ai: Fixed in 34bb14a. The workspace boundary now applies only to local file documents, so remote and other virtual schemes keep their existing validation behavior. I added regression coverage for an external vscode-remote document and a vscode-notebook-cell document, and aligned the README and setting description. Compile, client/server lint, and all client/server tests pass.

client/src/client.ts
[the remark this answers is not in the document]
QinXi-ai: Confirmed. In vscode-languageclient 10.1.0, provider.send delegates to the feature callback, which runs the configured middleware; the concrete open/close features then update their internal synchronization map after sending. The ESLint middleware also updates its local syncedDocuments map, so no additional map mutation is needed in refreshSyncedDocuments.

client/src/validator.ts
QinXi-ai: Updated as suggested: ValidateWorkspace is now a Pick of workspaceFolders, getConfiguration, and getWorkspaceFolder from typeof Workspace.

client/src/validator.ts
QinXi-ai: I initially broadened this beyond file, but that also catches vscode-notebook-cell and other virtual schemes, contrary to the PR scope of blocking local files while leaving virtual documents unchanged. Commit 34bb14a restores the file-only boundary and adds regression coverage for both vscode-remote and vscode-notebook-cell documents; the setting documentation now states that boundary explicitly.

QinXi-ai:
@dbaeumer Could you take another look at this PR when you have a chance? The earlier workspace-boundary and typing comments were addressed in 34bb14a, and the latest b3b5bf9 restores the legacy `autoFix` validation behavior identified in the follow-up review. The current head is b3b5bf9; the review threads are resolved.

I validated the current branch locally with TypeScript compilation, client/server lint, 25 client tests, and 5 server tests. The CLA check has passed. GitHub's “VS Code PR Check” has remained in progress since September 21, so I cannot claim that check is complete.

